PR #19, opened 2020-03-19
add exception 

Baranovsky1986:
add exceptions


criticus:
do we really need .xml file?

PR #24, opened 2020-04-02
Test download

Baranovsky1986:
Luyba. it`s not for push in project. I want to ask you check the unzipp_downloaded file. For smaller files of 500 mb it works, but for file 1.5 gb. the process stops when the file loads 1.1 gb,


Baranovsky1986:
old & not actual

PR #26, opened 2020-04-07
unzip_and_save_to_db

Vital-jan:


ratu/services/main.py
```diff
@@ -15,6 +15,42 @@ class Converter:
...
+def unzip_file(file_url, local_folder):
+        # getting zip file from file_url & extracting to local_folder
+        # must be call before parsing xml: Converter.unzip_file(FILE_URL, LOCAL_FOLDER)
+
+        def rename_xml_files (folder):
+            # analyse file list of directory <folder> & rename current unzipped files to uo.xml and fop.xml
+            # if filename not contains "uo" or "fop" string - file doesn`t rename.
+            # os module required
+            files = os.listdir (folder)
+            for el in files:
```
criticus, line 27: can you please rename `el`? It's not clear what it means

ratu/services/rfop_service.py
```diff
@@ -30,6 +30,35 @@ class RfopConverter(Converter):
...
+        state_id = None
+
+        for key in self.fop_status_array:
+            if (self.fop_status_array[key] == rec['STAN']):
+                # existing value of state
+                state_id = key
+
+        if (state_id == None):
+                # new value of Rfop.state
+                fopst = Fop_states (state = rec['STAN'][:100])
```
criticus, line 45: why is it `100`? Can you please add a comment to explain this line?

ratu/services/main.py
Vital-jan: +

ratu/services/rfop_service.py
Vital-jan: I think, for fix errors when string length in xml > string length in model for we must create constants, contains max string length and use them in models and in <save_to_db> module.

criticus: Please add the comment to the code

PR #30, opened 2020-04-13
deleted . from tests.py

IegorSoboliev:


ratu/tests.py
```diff
@@ -1,3 +1,3 @@
 from django.test import TestCase
 
-# Create your tests here.
```
Baranovsky1986, line 3: we need this dot

PR #32, opened 2020-04-13
Drf

Baranovsky1986:
on one documentation site, I found that I need to change these four files in this way. could you see if it’s right


criticus:
do you need to add all 186 files? I think you submitted psycopg2 by mistake

Baranovsky1986:
I cloned the project again and it was cloned with psycopg2. how about modified files?

data_converter/wsgi.py
```diff
@@ -11,6 +11,6 @@
 
 from django.core.wsgi import get_wsgi_application
 
-os.environ.setdefault("DJANGO_SETTINGS_MODULE", "data_converter.settings")
+os.environ.setdefault("DJANGO_SETTINGS_MODULE", "{{ data_converter }}.settings")
```
criticus, line 14: why do we need curly brackets {{}}?

manage.py
```diff
@@ -3,7 +3,7 @@
 import sys
 
 if __name__ == "__main__":
-    os.environ.setdefault("DJANGO_SETTINGS_MODULE", "data_converter.settings")
+    os.environ.setdefault("DJANGO_SETTINGS_MODULE", "{{ data_converter }}.settings")
```
criticus, line 6: same question here

Baranovsky1986:
not relevant

PR #38, opened 2020-04-17
Implemented storing kved.json to DB

IegorSoboliev:
Implemented storing kved.json to DB


ratu/migrations/0007_auto_20200417_1433.py
```diff
@@ -0,0 +1,28 @@
+# Generated by Django 3.0.5 on 2020-04-17 14:33
```
criticus, line 1: please double check django versions

ratu/services/kzed_service.py
```diff
@@ -0,0 +1,62 @@
...
+    # hope I will not need that
+    # dictionaries for keeping all model class objects
+    # section_dict={} 
+    # division_dict={}
+    # group_dict={}
+    # kzed_dict={}
+
+    #storing data to all tables       
+    def save_to_db(self, data):
+        data = data['sections'][0]
```
criticus, line 42: please add a comment to explain why we use `[0]`

PR #47, opened 2020-04-21
DRF pagination rfop & ruo

Vital-jan:


ratu/views.py
```diff
@@ -9,17 +9,50 @@
...
+            'data': data
+        }
+        return Response(data)
 
-class RuoView(APIView):
+class RuoView(GenericAPIView):
+    serializer_class = RuoSerializer
+    queryset = Ruo.objects.all()
+    pagination_class = CustomPagination
     def get(self, request):
```
Iva-khar, line 43: Why you define the same method in different classes?

criticus: +1 if you're reusing the same method it would make sense to move it to abstract class

ratu/views.py
```diff
@@ -9,17 +9,50 @@
...
+        page = self.paginate_queryset(queryset)
+        if page is not None:
+            serializer = self.get_serializer(page, many=True)
+            result = self.get_paginated_response(serializer.data)
+            data = result.data # pagination data
+        else:
+            serializer = self.get_serializer(queryset, many=True)
+            data = serializer.data
+        payload = {
+            'return_code': '0000',
```
criticus, line 33: what does this code mean?

Vital-jan: API response pagination

ratu/views.py
Vital-jan: ++

ratu/views.py
Iva-khar: What means '0000'?

Vital-jan: It`s built-in django value.

PR #50, opened 2020-04-22
Moving views and serializers under their own folders

Baranovsky1986:
changed project structure ( mkdir serializers & views )

### DIFF
--- a/ratu/serializers/ruo_serializers.py
+++ b/ratu/serializers/ruo_serializers.py
@@ -1,0 +1,10 @@
+from rest_framework import serializers
+
+class RuoSerializer(serializers.Serializer):
+    state = serializers.CharField(max_length=100)
+    kved = serializers.CharField()
+    name = serializers.CharField()
+    short_name = serializers.CharField()
+    edrpou = serializers.CharField()
+    address = serializers.CharField()
+    boss = serializers.CharField()
--- a/ratu/urls.py
+++ b/ratu/urls.py
@@ -1,13 +1,11 @@
 from django.urls import path
-from .views import RfopView
-from .views import RuoView
-from . import views
-from .views import RegionView, DistrictView, CityView, CitydistrictView, StreetView
+from ratu.views.rfop_views import RfopView
+from ratu.views.ruo_views import RuoView
+from ratu.views.ratu_views import RegionView, DistrictView, CityView, CitydistrictView, StreetView
 
 app_name = 'ratu.apps.RatuConfig'
 
 urlpatterns = [
-    path('', views.index, name='index'),
     path('rfop/', RfopView.as_view()),
     path('ruo/', RuoView.as_view()),
     path('region/', RegionView.as_view()),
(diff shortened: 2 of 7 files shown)


ratu/serializers/ruo_serializers.py
```diff
@@ -0,0 +1,10 @@
+from rest_framework import serializers
+
+class RuoSerializer(serializers.Serializer):
+    state = serializers.CharField(max_length=100)
```
criticus, line 4: is there any reason why we have max_length only for one field?

ratu/views/rfop_views.py
```diff
@@ -0,0 +1,12 @@
+from django.shortcuts import render
+from django.http import HttpResponse
+from rest_framework.response import Response
+from rest_framework.views import APIView
+from ratu.models.rfop_models import Rfop
+from ratu.serializers.rfop_serializers import RfopSerializer
+
+class RfopView(APIView):
+    def get(self, request):
+        rfop = Rfop.objects.all()[:5]
```
criticus, line 10: why is there limit 5?

ratu/views/ruo_views.py
```diff
@@ -0,0 +1,12 @@
+from django.shortcuts import render
+from django.http import HttpResponse
+from rest_framework.response import Response
+from rest_framework.views import APIView
+from ratu.models.ruo_models import Ruo
+from ratu.serializers.ruo_serializers import RuoSerializer
+
+class RuoView(APIView):
+    def get(self, request):
+        ruo = Ruo.objects.all()[:5]
```
criticus, line 10: same here

ratu/urls.py
```diff
@@ -1,13 +1,11 @@
...
+from ratu.views.ruo_views import RuoView
+from ratu.views.ratu_views import RegionView, DistrictView, CityView, CitydistrictView, StreetView
 
 app_name = 'ratu.apps.RatuConfig'
 
 urlpatterns = [
-    path('', views.index, name='index'),
     path('rfop/', RfopView.as_view()),
     path('ruo/', RuoView.as_view()),
     path('region/', RegionView.as_view()),
```
Iva-khar, line 11: Is it works with new destinations of 'RfopView' & others?

Baranovsky1986: yes it works

ratu/serializers/ruo_serializers.py
Baranovsky1986: it was in Vital request that you merged before my pr. I changed only (max_length=120) to (max_length=100)

ratu/views/rfop_views.py
Baranovsky1986: it was in Vital request that you merged before my pr.
as I understand it, we will change ".objects.all" to filter or anything else, so I think that now it doesn’t matter

ratu/views/ruo_views.py
Baranovsky1986: the same^

ratu/urls.py
Iva-khar: ok

PR #55, opened 2020-04-25
New download

Vital-jan:
compare filesize, download, unzip and rename files.

### DIFF
--- a/ratu/services/main.py
+++ b/ratu/services/main.py
@@ -1,32 +1,129 @@
 import codecs
 from collections import defaultdict
 from django.apps import apps
+import json
 import io
+import os
 import requests
 import sys
+import time
 from xml.etree.ElementTree import iterparse, XMLParser, tostring
 import xmltodict
 import zipfile
 
 class Converter:
-
-    FILE_URL=None
-    LOCAL_FOLDER=None
-    LOCAL_FILE_NAME=None
+    
+    UPDATE_FILE_NAME = "update.cfg"
+    LOCAL_FILE_NAME = None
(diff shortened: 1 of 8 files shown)


ratu/services/main.py
```diff
@@ -1,32 +1,132 @@
 import codecs
 from collections import defaultdict
 from django.apps import apps
+import json
 import io
+import os
+from os import path
```
criticus, line 7: why do you import `os` twice?

script.py
```diff
@@ -0,0 +1,14 @@
+from ratu.services.rfop_service import RfopConverter
```
criticus, line 1: what is script.py for?

ratu/services/main.py
```diff
@@ -1,32 +1,132 @@
...
 import requests
 import sys
+import time
```
criticus, line 10: where do you use `time`?

ratu/services/main.py
Vital-jan: fixed

script.py
Vital-jan: for one command run application with django_extention (your advice)

ratu/services/main.py
Vital-jan: already unused. fixed.

script.py
criticus: okay please add instructions to wiki how to run this script

Vital-jan: Ok. I`ll write this at wiki: python3 manage.py runscript script.py

Vital-jan: https://github.com/3v-workspace/Data_convertor/wiki/Django-Runscript

PR #56, opened 2020-04-28
Function get_first_word added to main.py

Anntroy:
Function get_first_word added to main.py

### DIFF
--- a/ratu/services/main.py
+++ b/ratu/services/main.py
@@ -23,6 +23,10 @@ class Converter:
     def __init__(self):
         return 
 
+        #geting a single uppercase word from some string
+    def get_first_word(self, string):
+        return string.upper().split()[0]        
+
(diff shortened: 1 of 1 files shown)


ratu/services/main.py
```diff
@@ -23,6 +23,10 @@ class Converter:
     def __init__(self):
         return 
 
+        #geting a single uppercase word from some string
+    def get_first_word(self, string):
+        return string.upper().split()[0]
```
Iva-khar, line 28: Maybe it's need to delete this method from koatuu_service.py?

Anntroy: Yes. I had removed it. It will appear in the next pull request with koatuu_service.py.

Iva-khar: You may change existing PR

PR #57, opened 2020-04-28
Ruo rfop  pagination

Baranovsky1986:
in this pr i & Vital worked together. We add  a pagination to ruo, rfop & ratu views


data_converter/urls.py
```diff
@@ -19,5 +19,5 @@
 
 urlpatterns = [
     path('api/', include('ratu.urls')),
-    path('admin/', admin.site.urls)
+    path('admin/', admin.site.urls),
```
Iva-khar, line 22: why?

PR #59, opened 2020-04-29
Adding category model in ratu_models, deleting clear_db function

IegorSoboliev:


ratu/models/ratu_models.py
```diff
@@ -1,43 +1,54 @@
...
+
     def __str__(self):
         return self.name
 
 class District(models.Model):
-    EMPTY_FIELD = 'empty field'
-    region = models.ForeignKey(Region, on_delete=models.CASCADE)
-    name = models.CharField(max_length=100)
+    region = models.ForeignKey(Region, on_delete=models.CASCADE, null=True)
+    name = models.CharField(max_length=100, null=True)
```
criticus, line 18: is there any example when District name is null?

ratu/models/ratu_models.py
```diff
@@ -1,43 +1,54 @@
...
 
+class Category(models.Model):
+    name = (models.CharField(max_length=5, unique=True, null=True))
+
+    def __str__(self):
+        return self.name
+
 class Region(models.Model):
-    name = models.CharField(max_length=30, unique=True)
+    name = models.CharField(max_length=30, unique=True, null=True)
```
criticus, line 10: when Region name can be null?

ratu/models/ratu_models.py
```diff
@@ -1,43 +1,54 @@
...
 
 class City(models.Model):
-    EMPTY_FIELD = 'empty field'
-    region = models.ForeignKey(Region, on_delete=models.CASCADE)
-    district = models.ForeignKey(District, on_delete=models.CASCADE)
-    name = models.CharField(max_length=100)
+    region = models.ForeignKey(Region, on_delete=models.CASCADE, null=True)
+    district = models.ForeignKey(District, on_delete=models.CASCADE, null=True)
+    category = models.ForeignKey(Category, on_delete=models.CASCADE, null=True)
+    name = models.CharField(max_length=100, null=True)
```
criticus, line 28: same here

ratu/models/ratu_models.py
```diff
@@ -1,43 +1,54 @@
...
-    region = models.ForeignKey(Region, on_delete=models.CASCADE)
-    district = models.ForeignKey(District, on_delete=models.CASCADE)
-    city = models.ForeignKey(City, on_delete=models.CASCADE)
-    name = models.CharField(max_length=100)
+    region = models.ForeignKey(Region, on_delete=models.CASCADE, null=True)
+    #let`s delete this attribute?
+    district = models.ForeignKey(District, on_delete=models.CASCADE, null=True)
+    category = models.ForeignKey(Category, on_delete=models.CASCADE, null=True)
+    city = models.ForeignKey(City, on_delete=models.CASCADE, null=True)
+    name = models.CharField(max_length=100, null=True)
```
criticus, line 40: and here

ratu/models/ratu_models.py
IegorSoboliev: Yes, in all cities

IegorSoboliev: And the same is possible in districts, citydistricts and streets models

ratu/services/main.py
```diff
@@ -135,12 +135,6 @@ def parse_file(self):
         with codecs.open(self.LOCAL_FOLDER + self.LOCAL_FILE_NAME, encoding="cp1251") as file:
             return xmltodict.parse(file.read())
     
-    def clear_db(self):
-        # clearing data base
-        for table in self.tables:
-            table.objects.all().delete()
-            print('Old data have deleted.')
```
Iva-khar, line 142: it cannot be deleted like that

ratu/models/ratu_models.py
criticus: in cities yes, but you added it in District :)
It's like region could be empty in district but name should have a value

IegorSoboliev: Sometimes no:). I will show you the reason

ratu/models/ratu_models.py
IegorSoboliev: Done

PR #65, opened 2020-05-01
Making DataOcean parent model

IegorSoboliev:


data_ocean/models/ratu_models.py
```diff
@@ -1,13 +1,14 @@
 from django.db import models
```
Iva-khar, line 1: Is it still need?

data_ocean/models/rfop_models.py
```diff
@@ -1,8 +1,9 @@
 from django.db import models
```
Iva-khar, line 1: the same

data_ocean/models/ruo_models.py
```diff
@@ -1,14 +1,15 @@
 from django.db import models
```
Iva-khar, line 1: the same

data_ocean/models/ratu_models.py
Iva-khar, line 1: Is it still need?

data_ocean/models/rfop_models.py
Iva-khar, line 1: the same

data_ocean/models/ruo_models.py
Iva-khar, line 1: the same

data_ocean/models/ratu_models.py
IegorSoboliev: Yes, because we still use functions of models

data_ocean/models/rfop_models.py
IegorSoboliev: Yes, because we still use functions of models

data_ocean/models/ruo_models.py
IegorSoboliev: Yes, because we still use functions of models

PR #68, opened 2020-05-03
DO 12 Implement API for model RUO

Iva-khar:
Implement API for Ruo with Chaining filters for all Model Fields like:
`/api/ruo/?id=''`
`/api/ruo/?state_id=''&address__icontains=''&name__icontains=''&kved__name=''` 
etc.


Iva-khar:
I think that method `get_queryset()` should be made abstract

data_ocean/views/ruo_views.py
```diff
@@ -1,13 +1,14 @@
...
-    queryset = Ruo.objects.all()
-    serializer = RuoSerializer(queryset, many=True)
-    pagination_class = CustomPagination
\ No newline at end of file
+    pagination_class = CustomPagination
+
+    def get_queryset(self):
+        queryset = Ruo.objects.all()
+        for key in self.request.query_params:
+            queryset = queryset.filter(**{key: self.request.query_params[key]})
```
criticus, line 13: I don't think it's a good idea to accept any parameters

PR #72, opened 2020-05-04
Adding Kved view with list and retrieve functions plus pagination

IegorSoboliev:


data_ocean/serializers/kved_serializers.py
```diff
@@ -1,7 +1,7 @@
 from rest_framework import serializers
 from data_ocean.models.kved_models import Kved
 
-class KvedSerializerWithModel(serializers.ModelSerializer):
+class KvedSerializer(serializers.ModelSerializer):
     class Meta:
         model = Kved
-        fields = ['code', 'name', 'group', 'division', 'section']
```
criticus, line 7: don't we want to show group, division, section?

data_ocean/views/kved_views.py
```diff
@@ -0,0 +1,23 @@
+from django.shortcuts import get_object_or_404
+from rest_framework import viewsets
+from rest_framework import generics
```
criticus, line 3: can you squash rest_framework imports into one line?

data_ocean/views/kved_views.py
```diff
@@ -0,0 +1,23 @@
...
+from rest_framework.response import Response
+from data_ocean.models.kved_models import Kved
+from data_ocean.serializers.kved_serializers import KvedSerializer
+from rest_framework.pagination import PageNumberPagination
```
criticus, line 7: order imports alphabetically please

data_ocean/views/kved_views.py
```diff
@@ -0,0 +1,23 @@
...
+
+class KvedView(viewsets.ReadOnlyModelViewSet, PageNumberPagination):
+    queryset = Kved.objects.all()
+    serializer_class = KvedSerializer
+
+    def list(self, request):
+        queryset = self.get_queryset()
+        results = self.paginate_queryset(queryset)
```
criticus, line 15: where do we set pagination parameters? offset, limit?

data_ocean/serializers/kved_serializers.py
IegorSoboliev: Usually, only these two parameters are shown

data_ocean/views/kved_views.py
IegorSoboliev: done

data_ocean/views/kved_views.py
IegorSoboliev: done with automatic sorting

data_ocean/views/kved_views.py
IegorSoboliev: My idea is to do that in settings for all classes

data_ocean/serializers/kved_serializers.py
criticus: usually? what do you mean? It's our first service :)

data_ocean/views/kved_views.py
criticus: ok, please do so

data_ocean/serializers/kved_serializers.py
IegorSoboliev: I mean in all other sources)

IegorSoboliev: And I hope this serializer will be used in other models. Two attributes are enough

data_ocean/views/kved_views.py
IegorSoboliev: done

data_ocean/serializers/kved_serializers.py
IegorSoboliev: I think the best solution is two have two serializers for Kved. Added new one with all attributes

data_converter/settings.py
```diff
@@ -40,6 +40,8 @@
     'django.contrib.messages',
     'django.contrib.staticfiles',
     'rest_framework',
+    'rest_framework.authtoken',
+    'rest_auth',
```
criticus, line 44: I guess this should be in the next task)) but it's ok

PR #78, opened 2020-05-06
Ratu service tables to lowercase and adapting to koatuu.json

Anntroy:
Deleted category items in fields like ratu_city.name.

### DIFF
--- a/data_ocean/services/ratu_service.py
+++ b/data_ocean/services/ratu_service.py
@@ -34,8 +34,8 @@ class RatuConverter(Converter):
     def rename (self, file):
         new_filename = ""
         if (file.upper().find('ATU') >= 0): new_filename = 'ratu.xml'
-        return new_filename  
-           
+        return new_filename
+    
     #creating dictionary & lists for registration items that had writed to db
     region_dict = {} # dictionary uses for keeping whole model class objects
     district_list = list() # lists use for keeping cells content
@@ -45,7 +45,7 @@ class RatuConverter(Converter):
     bulk_manager = BulkCreateManager(CHUNK_SIZE)
 
     #changing from records like: "волинська обл." to "волинська область"
-    def clear_region_name(self, region):
+    def clean_region_name(self, region):
         region = region.lower()
         region = re.sub(r"обл\.", "область", region)
         return region.strip()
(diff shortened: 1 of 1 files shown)


data_ocean/services/ratu_service.py
```diff
@@ -32,15 +34,41 @@ class RatuConverter(Converter):
...
+    #creating dictionary & lists for registration items that had writed to db
     region_dict = {} # dictionary uses for keeping whole model class objects
     district_list = list() # lists use for keeping cells content
     city_list = list()
     citydistrict_list = list()
 
     bulk_manager = BulkCreateManager(CHUNK_SIZE)
+
+    #changing from records like: "волинська обл." to "волинська область"
+    def clear_region_name(self, region):
```
Baranovsky1986, line 48: One trivial comment. Why the name of this method is different from those below (clear instead of clean)?

Anntroy: It's a mistake. I'll change it

data_ocean/services/ratu_service.py
```diff
@@ -32,15 +34,41 @@ class RatuConverter(Converter):
...
+        region = region.lower()
+        region = re.sub(r"обл\.", "область", region)
+        return region.strip()
+
+    #changing from records like: "донецький р-н" to "донецький район", "р.райони вінницької області" to "райони вінницької області", "райони міста Київ" to "райони м.Київ"
+    def clean_district_name(self, district):
+        district = district.lower()
+        district = re.sub(r"р-н", "район", district)
+        district = re.sub(r"р\.", "", district)
+        district = re.sub(r"райони міста", "райони м.", district)
```
IegorSoboliev, line 58: is that correct renaming "райони м."?

data_ocean/services/ratu_service.py
```diff
@@ -127,14 +156,14 @@ def save_to_citydistrict_table(self, record, region, district, city):
...
-    def save_to_street_table(self, record, region, district, city, citydistrict):    
+    def save_to_street_table(self, record, region, district, city, citydistrict):
         if record['STREET_NAME']:
             street = Street(
                 region=region, 
                 district=district,
                 city=city,
                 citydistrict=citydistrict,
-                name=record['STREET_NAME']
+                name=record['STREET_NAME'].lower()
```
IegorSoboliev, line 166: Do we really need to have a street`s and others names lower case? We are going to show all  these values to the public))

criticus: it's better to store them in one style in the db for faster search. We will modify it on the frontend for UI

criticus:
looks ok to me

PR #80, opened 2020-05-06
Ratu list retrieve

Baranovsky1986:
add list & retrieve method for ratu_views


IegorSoboliev:
@criticus looks like we can make an abstract View because all views are similar now

criticus:
@IegorSoboliev I don't think it worth to make it abstract now. Let's see how it goes further

PR #81, opened 2020-05-06
get_id_rfop

Baranovsky1986:
modify rfop serializers to ModelSerializer & add list & retriever to rfopviews


update.cfg
```diff
@@ -1 +1 @@
-{"https://data.gov.ua/dataset/75e57837-128b-49e1-a007-5e7dfa7bf6af/resource/e21a1e57-051c-46ea-9c8e-8f30de7d863d/download/": 3326459, "https://data.gov.ua/dataset/b244f35a-e50a-4a80-b704-032c42ba8142/resource/06bbccbd-e19c-40d5-9e18-447b110c0b4c/download/": 448365062}
\ No newline at end of file
+{"https://data.gov.ua/dataset/75e57837-128b-49e1-a007-5e7dfa7bf6af/resource/e21a1e57-051c-46ea-9c8e-8f30de7d863d/download/": 3326571, "https://data.gov.ua/dataset/b244f35a-e50a-4a80-b704-032c42ba8142/resource/06bbccbd-e19c-40d5-9e18-447b110c0b4c/download/": 448921115}
```
criticus, line 1: why do you need to update .cfg?

IegorSoboliev: I think we should build a new model for Rfop first

Baranovsky1986: I did`t do this, I did`t touch this file. I saw that the deleted and added line are the same

criticus: @Baranovsky1986 if you don't want these changes then revert them

Baranovsky1986:
it has conflict

PR #83, opened 2020-05-07
Save_to_district_table  and save_to_city_table in koatuu_service

Anntroy:


data_ocean/services/koatuu_service.py
```diff
@@ -1,6 +1,7 @@
 import config
 from data_ocean.services.main import Converter
 from data_ocean.models.ratu_models import Region, District, City, Citydistrict
+import re
```
criticus, line 4: is it used?

PR #84, opened 2020-05-07
Ruo views

Iva-khar:


data_ocean/services/ruo_service.py
```diff
@@ -51,7 +51,6 @@ def save_to_db(self, record):
         state=self.save_to_state_table(record)
         kved=self.get_kved_from_DB(record, 'NAME')
         self.save_to_ruo_table(record, state, kved)
-        # self.save_to_founders_table(record, ruo)
```
criticus, line 54: why is this commented out?

Iva-khar: It Not need with Bulkcreate

criticus: then please delete it

PR #85, opened 2020-05-07
(WIP)Admin user

Baranovsky1986:
Create User Authentication, Custom User Model (now we can create a new user account, login, registration, logout, and view all users via endpoints.)


api/admin.py
```diff
@@ -0,0 +1,3 @@
+from django.contrib import admin
```
criticus, line 1: why do we need a separate app API?

data_converter/urls.py
```diff
@@ -20,4 +20,5 @@
 urlpatterns = [
     path('api/', include('data_ocean.urls')),
     path('admin/', admin.site.urls),
+    path('api/v1/', include('api.urls')),
```
criticus, line 23: we already have `api` (line 21). Why are you adding new `api`?

users/admin.py
```diff
@@ -0,0 +1,15 @@
+from django.contrib import admin
+from django.contrib.auth import get_user_model
+from django.contrib.auth.admin import UserAdmin
+
+from .forms import CustomUserCreationForm, CustomUserChangeForm
+from .models import CustomUser
+
+class CustomUserAdmin(UserAdmin):
+    add_form = CustomUserCreationForm
+    form = CustomUserChangeForm
```
criticus, line 10: we do not need form as of now

users/models.py
```diff
@@ -0,0 +1,10 @@
+from django.db import models
+
+from django.contrib.auth.models import AbstractUser
+from django.db import models
+
+class CustomUser(AbstractUser):
```
criticus, line 6: please name it DataOceanUser

PR #86, opened 2020-05-08
Common models for Status, Authority and Taxpayer

IegorSoboliev:


data_ocean/models/common_models.py
```diff
@@ -0,0 +1,12 @@
+from django.db import models
+from data_ocean.models.main import DataOceanModel
+
+class Status(DataOceanModel):
+    name = models.CharField(max_length=100, unique=True)
+
+class Authority(DataOceanModel):
+    name = models.CharField(max_length=500, unique=True)
+    code = models.CharField(max_length=10, unique=True, null=True)
```
criticus, line 9: what is code? are their predefined somewhere?

Iva-khar: fo sample:
```
AUTHORITY_NAME                  ДПI У ЦЕНТРАЛЬНОМУ РАЙОНI М.ДНIПРА ГУ ДФС У ДНIПРОПЕТРОВСЬКIЙ ОБЛАСТI
                 AUTHORITY_CODE                  39533006
```

IegorSoboliev: As far as I can see code here is a EDRPO of state body

IegorSoboliev: An example https://dtkt.com.ua/show/0sid0329.html

IegorSoboliev: All such data should be there http://lg.ukrstat.gov.ua/inform_z_edrpou.php.htm

data_ocean/models/common_models.py
```diff
@@ -0,0 +1,12 @@
...
+
+class TaxpayerType(DataOceanModel):
+    name = models.CharField(max_length=200, unique=True)
```
Iva-khar, line 12: ABC order )

PR #92, opened 2020-05-12
Simple_history for Fop, FopToKved, ExchangeData models

IegorSoboliev:


data_ocean/admin.py
```diff
@@ -1,9 +1,12 @@
 from django.contrib import admin
-from data_ocean.models.rfop_models import Rfop
+from data_ocean.models.rfop_models import Rfop, Fop, FopToKved, ExchangeData
 from data_ocean.models.ratu_models import Region, District, City, Citydistrict, Street
 from data_ocean.models.kved_models import Kved, Group, Division, Section
 
 admin.site.register(Rfop)
+admin.site.register(Fop)
+admin.site.register(FopToKved)
```
criticus, line 8: is this needed as a separate admin page?

criticus:
per our conversation @IegorSoboliev is going to remove FopToKved and ExchangeData from Historical records

data_ocean/admin.py
IegorSoboliev: removed

data_converter/settings.py
```diff
@@ -42,7 +42,8 @@
     'rest_framework',
     'rest_framework.authtoken',
     'rest_auth',
-    'django_extensions'
+    'django_extensions',
+    'simple_history',
```
criticus, line 46: does it need to be in requirements?

IegorSoboliev: Sorry, I didn`t understand you. No, that`s enough

IegorSoboliev: it is here already)

PR #94, opened 2020-05-14
History for Company, CompanyDetail

IegorSoboliev:


data_ocean/migrations/0046_auto_20200508_1730.py
```diff
@@ -7,7 +7,7 @@
 class Migration(migrations.Migration):
 
     dependencies = [
-        ('data_ocean', '0045_auto_20200508_1422'),
+        ('data_ocean', '0046_auto_20200513_1407'),
```
criticus, line 10: have you fixed migrations dependency already?

data_ocean/models/company_models.py
```diff
@@ -46,6 +48,7 @@ class CompanyDetail(DataOceanModel):
     terminated_info = models.CharField(max_length=100, null=True)
     termination_cancel_info = models.CharField(max_length=100, null=True)
     vp_dates = models.CharField(max_length=100, null=True)
+    history = HistoricalRecords()
```
criticus, line 51: why do we need historical records for this model?

PR #96, opened 2020-05-15
admin_users

Baranovsky1986:
Create a new user account, login, logout, and view all users via endpoints.


users/views.py
```diff
@@ -0,0 +1,10 @@
+from django.shortcuts import render
+
+from rest_framework import generics
+
+from . import models
+from . import serializers
+
+class UserListView(generics.ListAPIView):
```
criticus, line 8: is there a way to get a specific user?

Baranovsky1986: Do you mean to get user by id?

criticus: yes

Baranovsky1986: it can be added, but at first it would be nice to merge it

Baranovsky1986: in order to avoid conflicts at this stage, as a basic application that can then be improved

PR #100, opened 2020-05-20
merge

Bevcuk:


data_converter/settings_local.base.py
```diff
@@ -0,0 +1,22 @@
+
+import os
+BASE_DIR = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
+# DATABASE_CONFIG = {
+#     'ENGINE': 'django.db.backends.sqlite3',
```
Anntroy, line 5: Why sqlite3, and not Psql?

location_register/converter/koatuu.py
```diff
@@ -1,6 +1,6 @@
-import config
-from data_ocean.services.main import Converter
-from data_ocean.models.ratu_models import Region, District, City, Citydistrict
+from data_ocean.converter import Converter
+from location_register.models import Region, District, City, Citydistrict
+
 
 class KoatuuConverter(Converter):
```
Anntroy, line 6: Can you, please, uncomment lines: 95, 101, 105, 106, 110. I sow now that I left them commented by error.

data_converter/settings_local.base.py
Bevcuk: it is only example file

location_register/converter/koatuu.py
Bevcuk: done

data_converter/settings_local.base.py
Anntroy: ok

location_register/converter/koatuu.py
Anntroy: Thanks

Anntroy: Thanks

PR #101, opened 2020-05-21
Make service for fop

IegorSoboliev:


data_ocean/services/fop_service.py
```diff
@@ -0,0 +1,119 @@
+import config
```
Bevcuk, line 1: why in data_ocean/services?

PR #102, opened 2020-05-21
Models and API for register table

Anntroy:


data_ocean/register_models.py
```diff
@@ -0,0 +1,21 @@
+from django.db import models
```
Bevcuk, line 1: why do you rename file to register_model?

Anntroy: I didn't rename, I've added new files.
But, I suppose, you suggest me to add my models in the existing file models.

Bevcuk: yes

PR #105, opened 2020-05-21
New parser for xml data sources

Iva-khar:
New parser for xml data sources

### DIFF
--- a/data_ocean/converter.py
+++ b/data_ocean/converter.py
@@ -3,6 +3,7 @@ from collections import defaultdict
 from django.apps import apps
 import json
 import io
+from lxml import etree
 import os
 import requests
 import sys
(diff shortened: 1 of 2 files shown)


data_ocean/converter.py
```diff
@@ -222,7 +223,22 @@ def process(self):
         except:
             None
         print('All the records have been rewritten.')
-
+    
+    def process_full(self): # It's temporary method name, in the future this 'process' will be one       
+        i = 0
+        records = etree.Element('RECORDS')
+        for _, elem in etree.iterparse(self.LOCAL_FOLDER + self.LOCAL_FILE_NAME, tag = 'SUBJECT'):
```
Anntroy, line 230: "_," what does it mean?

data_ocean/converter.py
```diff
@@ -222,7 +223,22 @@ def process(self):
         except:
             None
         print('All the records have been rewritten.')
-
+    
+    def process_full(self): # It's temporary method name, in the future this 'process' will be one       
+        i = 0
+        records = etree.Element('RECORDS')
+        for _, elem in etree.iterparse(self.LOCAL_FOLDER + self.LOCAL_FILE_NAME, tag = 'SUBJECT'):           
+            if len(records) < self.CHUNK_SIZE:
```
Anntroy, line 231: Where do you define CHUNK_SIZE?

Iva-khar: It will in services

data_ocean/converter.py
Iva-khar: some required but no realy need 
variable

PR #106, opened 2020-05-21
Common functions for Converter

IegorSoboliev:


data_ocean/converter.py
```diff
@@ -225,6 +207,58 @@ def process(self):
 
     print('Converter has imported.')
 
+    #lets have all supporting functions in the end of the class
+    def get_first_word(self, string, upper=False):
```
Anntroy, line 211: I don't use this method anymore. Maybe, you can gather it with your get_other_word method.

data_ocean/converter.py
```diff
@@ -225,6 +207,58 @@ def process(self):
...
+        return string.upper().split()[0] if upper else string.split()[0]
+
+    def get_other_words(self, string):
+        return string.split()[1:]
+
+    def get_kved_from_DB(self, kved_code_from_record):
+        # verifying kved
+        empty_kved = Kved.objects.get(code='EMP')
+        kved_dict = {}    
+        for kved in Kved.objects.all():
```
Anntroy, line 221: You can store Kved.objects.all() in some value. So we'll be making just one request to db and not for every kved. Luyba says it saves running time for our program

data_ocean/converter.py
```diff
@@ -225,6 +207,58 @@ def process(self):
...
+            kved_dict[kved.code]=kved
+        if kved_code_from_record in kved_dict:
+            return kved_dict[kved_code_from_record]
+        else:
+            print(f"This kved value is not valid")
+            return empty_kved
+    
+    def get_or_save_status(self, status_from_record):
+        status_dict = {}
+        for status in Status.objects.all():
```
Anntroy, line 231: Consider creating some value to store Status.objects.all()

data_ocean/converter.py
```diff
@@ -225,6 +207,58 @@ def process(self):
...
+            status_dict[status.name] = status
+        if not status_from_record in status_dict:
+            new_status = Status(name=status_from_record)
+            new_status.save()
+            return new_status
+        else:
+            return status_dict[status_from_record]
+
+    #later we should add a code if exists
+    def get_or_save_authority(self, authority_from_record):
```
Anntroy, line 241: You have three methods, which are almost the same. Consider to put them in the same method and add one more argument for Model_name

data_ocean/converter.py
IegorSoboliev: No. I use both separately

data_ocean/converter.py
IegorSoboliev: make sense, thank you

data_ocean/converter.py
IegorSoboliev: @criticus I`d like to hear your opinion

data_ocean/converter.py
IegorSoboliev: made universal function

Anntroy: ok

IegorSoboliev: In the end, I split into three functions again - for more readable code

PR #107, opened 2020-05-22
New model for registers table with api_adress fields

Anntroy:


data_ocean/models.py
```diff
@@ -24,3 +24,23 @@ class Authority(DataOceanModel):
...
+    url_adress = models.URLField(max_length=500)
+    api_adress = models.URLField(max_length=500, null=True)
+    source = models.ForeignKey(Source, on_delete=models.CASCADE)
+
+
+class Register(DataOceanModel):
+    name = models.CharField(max_length=500)
+    source = models.ForeignKey(Source, on_delete=models.CASCADE)
+    url_adress = models.ForeignKey(SourceAdress, on_delete=models.CASCADE)
+    api_adress = models.ForeignKey(SourceAdress, on_delete=models.CASCADE)
```
Anntroy, line 43: Is it ok, to have two foreign keys from the same table, but different fields?

data_ocean/models.py
```diff
@@ -24,3 +24,23 @@ class Authority(DataOceanModel):
...
 class TaxpayerType(DataOceanModel):
     name = models.CharField(max_length=200, unique=True)
+
+
+class Source(DataOceanModel):
+    name = models.CharField(max_length=300)
+
+
+class SourceAdress(DataOceanModel):
+    url_adress = models.URLField(max_length=500)
```
criticus, line 34: address with double d

data_ocean/models.py
```diff
@@ -24,3 +24,23 @@ class Authority(DataOceanModel):
 
 class TaxpayerType(DataOceanModel):
     name = models.CharField(max_length=200, unique=True)
+
+
+class Source(DataOceanModel):
+    name = models.CharField(max_length=300)
```
criticus, line 30: why is this a separate model?

data_ocean/models.py
criticus: it's okay but I don't see a point

data_ocean/models.py
Anntroy: ok

data_ocean/models.py
Anntroy: Because we can have a few URL addresses in the same source.

data_ocean/models.py
Anntroy: It was just my doubt

data_ocean/models.py
Bevcuk: You can do something like this:

class SourceAddress(DataOceanModel):
   name = models.CharField(max_length=300)
    url_address = models.URLField(max_length=500)
    api_address = models.URLField(max_length=500, null=True)

class Register(DataOceanModel):
    name = models.CharField(max_length=500)
    source = models.ForeignKey(Source, on_delete=models.CASCADE)

    def __str__(self):
        return self.name 
or 
class Register(DataOceanModel):
    name = models.CharField(max_length=500)
    source =models.CharField(max_length=300)
    url_address = models.URLField(max_length=500)
    api_address = models.URLField(max_length=500, null=True)

    def __str__(self):
        return self.name

Anntroy: So it's not necessary to do it separately?

Bevcuk: if we have a lot of address for one source, then just delete url_adress and api_adress with register

Anntroy: ok

data_ocean/serializers.py
```diff
@@ -19,4 +19,10 @@ class TaxpayerTypeSerializer(serializers.ModelSerializer):
     name = serializers.CharField(max_length=200)
```
Anntroy, line 19: Is this field necesary in here?

business_register/converter/kved.py
```diff
@@ -1,10 +1,10 @@
 from business_register.models.kved_models import Section, Division, Group, Kved
 from data_ocean.converter import Converter
-
+from data_ocean.models import Register
 
 class KvedConverter(Converter):
     LOCAL_FILE_NAME = "kved.json"
-    DATASET_ID = "e1afb81c-70e4-4009-96a0-b240c36e4603"
+    API_ADDRESS_FOR_DATASET = Register.objects.get(id=2).api_address
```
criticus, line 7: what is `2`?

business_register/converter/rfop.py
```diff
@@ -2,11 +2,12 @@
 from business_register.models.rfop_models import Rfop
 from business_register.models.ruo_models import State
 from data_ocean.converter import Converter, BulkCreateManager
+from data_ocean.models import Register
 
 
 class RfopConverter(Converter):
     LOCAL_FILE_NAME = "fop.xml"
-    DATASET_ID = "1c7f3815-3259-45e0-bdf1-64dca07ddc10"
+    API_ADDRESS_FOR_DATASET = Register.objects.get(id=1).api_address
```
criticus, line 10: what is `1`?

data_ocean/fixtures/register.json
```diff
@@ -0,0 +1,46 @@
+[
+    {
+      "model": "data_ocean.register",
+      "id": 501,
```
criticus, line 4: are these ids different from ids in models?

Anntroy: Yes, they are. In my model, I have from 1 to 4, which are given automatically by Django

business_register/converter/kved.py
Anntroy: It's the id number in my register table for Kved register.
Maybe it's better to search by name like: (name="Класифікація видів економічної діяльності (КВЕД)")?

business_register/converter/rfop.py
Anntroy: It's the id number in my register table for Roufull register.

data_ocean/fixtures/register.json
criticus: 1-4 could be different in the production database. I think it's better to get them by names rather than ids

business_register/converter/kved.py
Anntroy: Changed

data_ocean/fixtures/register.json
Anntroy: I've change all

PR #110, opened 2020-05-26
Add category_id values to city and citydistrict tables

Anntroy:


location_register/converter/ratu.py
```diff
@@ -61,12 +62,26 @@ def clean_city_or_citydistrict_name(self, city):
         city = re.sub(r"с\.|м\.|смт\.|сщ\.|с/рада\.|сщ/рада\.|р\.", "", city)
         return city.strip()
 
+    # function to compare data records with category types
+    def category(self, name):
```
IegorSoboliev, line 66: maybe, you can name this function "format_category_name"

Anntroy: Done

PR #111, opened 2020-05-26
proxy registration

Bevcuk:


users/serializers.py
```diff
@@ -1,8 +1,35 @@
...
 
 
-class UserSerializer(serializers.ModelSerializer):
-    class Meta:
-        model = models.DataOceanUser
-        fields = ('email', 'username', )
\ No newline at end of file
+class RegisterSerializer(serializers.Serializer):
+    email = serializers.EmailField(required=True)
+    password1 = serializers.CharField(write_only=True)
```
criticus, line 11: why is it called `password1`?

PR #112, opened 2020-05-29
Register table and API

Anntroy:
I've deleted source file ratu and koatuu. It shows that koatuu file was deleted.
Old PR 107


Anntroy:
Old PR 107 with deleted source files

PR #113, opened 2020-05-31
Register models api. Added last_update field, with update information from source

Anntroy:


data_ocean/models.py
```diff
@@ -31,6 +31,7 @@ class Register(DataOceanModel):
     source_name = models.CharField(max_length=300)
     url_address = models.URLField(max_length=500)
     api_address = models.URLField(max_length=500, null=True)
+    last_update = models.DateTimeField(auto_now_add=True, null=True)
```
roman-tiukh, line 34: maybe need dataset id? like source_register_id

data_ocean/models.py
```diff
@@ -31,6 +31,7 @@ class Register(DataOceanModel):
     source_name = models.CharField(max_length=300)
```
roman-tiukh, line 31: i think "name" must be unique

location_register/converter/koatuu.py
```diff
@@ -7,7 +7,7 @@ class KoatuuConverter(Converter):
 
     #paths for the local souce file
     LOCAL_FILE_NAME = "koatuu.json"
-    API_ADDRESS_FOR_DATASET = Register.objects.get(name="Класифікатор об’єктів адміністративно-територіального устрою України (КОАТУУ)").api_address
+    API_ADDRESS_FOR_DATASET = Register.objects.get(name="Класифікатор об’єктів адміністративно-територіального устрою України").api_address
```
roman-tiukh, line 10: get by id its may be better than by name, and why name not unique?

data_ocean/models.py
roman-tiukh, line 34: what is last_update? and why auto_now_add?

roman-tiukh: we need to add docstring or help_text or verbose_name or just comment to understand what it is.
maybe rename field to source_last_update?

data_ocean/models.py
Anntroy: Do you think short_name must be added as well?

data_ocean/models.py
Anntroy: Fixed source_register_id as URLField

PR #122, opened 2020-06-05
Parser, service, serializer and viewset for drv register

IegorSoboliev:


location_register/converter/drv.py
```diff
@@ -0,0 +1,191 @@
...
+from zeep.helpers import serialize_object
+
+from data_ocean.converter import Converter
+from location_register.models.drv_models import (DrvAto, DrvBuilding,
+                                                 DrvCouncil, DrvDistrict,
+                                                 DrvRegion, DrvStreet, ZipCode)
+
+
+class DrvConverter(Converter):
+    WSDL_URL = 'https://www.drv.gov.ua/ords/svc/personal/API/Opendata'
```
criticus, line 14: @Anntroy should this be in the all registers table in the db?

location_register/converter/drv.py
```diff
@@ -0,0 +1,191 @@
...
+    SETTINGS = Settings(strict=False, xml_huge_tree=True)
+    # dictionaries and lists for storing all objects from DB
+    all_regions_dict = {}
+    all_atos_dict = {}
+    all_zipcodes_dict = {}
+
+    # deleting 'м.', 'смт', 'с.', 'вул.' from string 
+    def clean_name(self, name):
```
criticus, line 22: consider moving this to utils, this is not the first time we use similar functionality

location_register/migrations/0010_auto_20200605_1305.py
```diff
@@ -0,0 +1,18 @@
+# Generated by Django 2.0.9 on 2020-06-05 13:05
```
criticus, line 1: consider squashing migrations into one

location_register/converter/drv.py
IegorSoboliev: I plan to establish common utils files in the next PR

location_register/converter/drv.py
IegorSoboliev: It depends on the purpose of having this table

location_register/migrations/0010_auto_20200605_1305.py
IegorSoboliev: done

location_register/converter/drv.py
Anntroy: No. It's a new register, we need to add this information to the existing Register table.

location_register/models/drv_models.py
```diff
@@ -4,11 +4,11 @@
 
 
 class DrvRegion(DataOceanModel):
```
Anntroy, line 6: You have 8 migration files in the same PR!

location_register/converter/drv.py
Anntroy: @IegorSoboliev I can't find it's register_id. I need this information to add to Register table.

IegorSoboliev: @Anntroy what is register_id?

PR #126, opened 2020-06-08
Custom User Model - DataOceanUser.

Anntroy:
Created Custom User Model to login using email address instead of username.


users/models.py
```diff
@@ -1,7 +1,47 @@
...
 from django.contrib.auth.models import AbstractUser
+from django.contrib.auth.base_user import BaseUserManager
+from django.utils.translation import ugettext_lazy as _
+
+class DataOceanUserManager(BaseUserManager):
+    """
+    Custom user model manager where email is the unique identifiers
+    for authentication instead of usernames.
+    """
+    def create_user(self, email, password, **extra_fields):
```
roman-tiukh, line 11: where is first_name? last_name?

Anntroy: **extra_fields include: 'last_login', 'is_superuser', 'first_name', 'last_name', 'is_staff', 'is_active', 'date_joined', 'groups', 'user_permissions'. You can see it in migration file: users/migrations/0001_initial.py.

PR #128, opened 2020-06-09
add system-wide cache

Baranovsky1986:
add Redis as system-wide cache


data_converter/settings.py
```diff
@@ -157,3 +157,14 @@
...
+CACHES = {
+    'default': {
+        'BACKEND': 'django_redis.cache.RedisCache',
+        'LOCATION': 'redis://127.0.0.1:6379/',
+        'OPTIONS': {
+            # 'PASSWORD': 'XXXXXXXXX',
+            'CLIENT_CLASS': 'django_redis.client.DefaultClient',
+        }
+    }
+}
```
roman-tiukh, line 170:
```suggestion
}

```

Baranovsky1986: I fixed it

requirements.txt
```diff
@@ -18,6 +18,7 @@ pylint==2.4.4
 pylint-django==2.0.15
 pylint-plugin-utils==0.6
 pytz==2019.3
+Redis==4.0.9
```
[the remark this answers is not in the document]
Baranovsky1986: We will discus about it. but now i need to merge with this version

PR #154, opened 2020-06-16
Added more checks for lacking data

IegorSoboliev:


business_register/converter/additional_uo.py
```diff
@@ -32,27 +32,33 @@ def parse_file(self):
     def add_founders_info(self, records):
         j = 0
```
roman-tiukh, line 33: unused variable

PR #157, opened 2020-06-16
Refactored converting equity value to float

IegorSoboliev:


business_register/converter/additional_uo.py
```diff
@@ -61,7 +61,8 @@ def add_founders_info(self, records):
             if founder_equity_info:
                 founder_equity = founder_equity_info[0].text
             if founder_equity:
-                founder_equity = float(founder_equity.replace(',', '.'))
+                founder_equity = founder_equity.replace(',', '.')
+                founder_equity = float(founder_equity)
```
criticus, line 65: have you tested if it works as expected?

roman-tiukh: what reason of this change?

business_register/models/company_models.py
```diff
@@ -17,7 +17,7 @@ class Company(DataOceanModel): #constraint for not null in both name & short_nam
     name = models.CharField(max_length=500, null=True)
     short_name = models.CharField(max_length=500, null=True)
     company_type = models.ForeignKey(CompanyType, on_delete=models.CASCADE)
-    edrpou = models.CharField(max_length=260)
+    edrpou = models.CharField(max_length=260, db_index=True)
```
roman-tiukh, line 20: is "edrpou" have max_lenght 260? why?

business_register/converter/additional_uo.py
roman-tiukh: this is not a solution of problem

PR #168, opened 2020-06-17
Celery-beat admin use for periodic tasks

Anntroy:


location_register/tasks.py
```diff
@@ -0,0 +1,27 @@
...
+def fill_in_ratu_table():
+    RatuConverter().process()
+
+@shared_task
+def fill_in_koatuu():
+    with open(LOCAL_FILE_KOATUU) as json_file:
+        data = json.load(json_file)
+    KoatuuConverter().save_to_db(data)
+
+@shared_task
```
roman-tiukh, line 23: do you use pylint? 2 blank lines betwen blocks.

data_converter/settings_local.base.py
```diff
@@ -76,3 +76,11 @@
 # business_register/converter/uo.py
 UO_CHUNK_SIZE = 100
 
+# celery settings
+
+CELERY_BROKER_URL = 'redis://localhost:6379' # redis://:password@hostname:port/db_number
```
roman-tiukh, line 81: redis backend need to use db

roman-tiukh: Because we have redis as cache backend for django.
He must use db name too. @Baranovsky1986

roman-tiukh: look at template in this line 
# redis://:password@hostname:port/db_number

location_register/tasks.py
Anntroy: Yes, I'm using pylint. In this case, there was no error shown by pylint.

data_converter/settings_local.base.py
Anntroy: Added 0 db_number. The default one

data_converter/settings_local.base.py
```diff
@@ -76,3 +76,11 @@
 # business_register/converter/uo.py
 UO_CHUNK_SIZE = 100
 
+# celery settings
+
+CELERY_BROKER_URL = 'redis://localhost:6379/0' # redis://:password@hostname:port/db_number
+CELERY_BROKER_TRANSPORT_OPTIONS = {'visibility_timeout': 3600}
+CELERY_RESULT_BACKEND = 'redis://localhost:6379/0'
```
roman-tiukh, line 83: why db is 0?

roman-tiukh: and change db in CHACHE setting

Anntroy: It's the default value. See here: https://docs.celeryproject.org/en/stable/getting-started/brokers/redis.html. In Configuration... all fields after the scheme are optional and will default to localhost on port 6379, using database 0.

roman-tiukh: This is the problem. No need to use default values. In order to avoid data conflicts, you need to set db = 1 for the CACHE option, and db = 2 for celery

data_converter/settings_local.base.py
```diff
@@ -58,7 +58,7 @@
     'default': {
         # REDIS cache configs
         # 'BACKEND': 'django_redis.cache.RedisCache',
-        # 'LOCATION': 'redis://127.0.0.1:6379/',
+        # 'LOCATION': 'redis://127.0.0.1:6379/0',
```
roman-tiukh, line 61: why you use same db for cache and celery???

data_converter/settings_local.base.py
Anntroy: Ok. I didn't know it

data_converter/settings_local.base.py
Anntroy: Changed

PR #173, opened 2020-06-18
historical_founder_service

Iva-khar:


business_register/converter/history.py
```diff
@@ -71,3 +71,38 @@ def save_to_db(self, records):
...
+            try:
+                company_exists = Company.objects.filter(edrpou=edrpou).first()
+                founder.company = company_exists
+            except AttributeError:
+                continue
+            try:
+                founder.id = Signer.objects.filter(company=company_exists).first().id
+            except AttributeError:
+                founder.id = 0 #for changed records that can't be assigned to existing company
+            founder.history_type = '~'
```
criticus, line 103: what does tilde mean here?

roman-tiukh: This is type `update`.
I think you need provide constants for history types.

Iva-khar: This Class for only one time using

roman-tiukh: add comment at least

PR #174, opened 2020-06-18
Add fields to DataOceanUser

PyDev777:


users/models.py
```diff
@@ -39,6 +39,9 @@ def create_superuser(self, email, password, **extra_fields):
 class DataOceanUser(AbstractUser):
     username = None
     email = models.EmailField(_('email address'), unique=True)
+    organization = models.CharField(max_length=255, default='')
+    position = models.CharField(max_length=150, default='')
```
roman-tiukh, line 43: blank = True

roman-tiukh: what is position?

users/models.py
```diff
@@ -39,6 +39,9 @@ def create_superuser(self, email, password, **extra_fields):
...
+    date_of_birth = models.DateField(null=True, blank=True)
```
roman-tiukh, line 44: default = None

users/models.py
PyDev777: Fixed

users/models.py
PyDev777: Fixed

PR #181, opened 2020-06-22
missed requirements

roman-tiukh:


requirements.txt
```diff
@@ -1,8 +1,11 @@
+amqp==2.6.0
 appdirs==1.4.4
 asgiref==3.2.7
 astroid==2.4.2
 attrs==19.3.0
+billiard==3.6.3.0
 cached-property==1.5.1
+celery==4.4.5
```
Anntroy, line 8: In my case, django-celery-beat is working without celery configured separately. That's why I didn't include it. But I suppose, you have generated this file automatically. Please, show us how to check this.

roman-tiukh: django-celery-beat depend of celery
image

Anntroy: ok

PR #185, opened 2020-06-23
Do 138 add CustomPagination 

Baranovsky1986:
add CustomPagination

### DIFF
--- a/data_converter/pagination.py
+++ b/data_converter/pagination.py
@@ -2,11 +2,8 @@ import math
 from rest_framework.pagination import PageNumberPagination
 from rest_framework.response import Response
 
-DEFAULT_PAGE = 1
-
 
 class CustomPagination(PageNumberPagination):
-    page = DEFAULT_PAGE
     page_size = 10
     page_size_query_param = 'page_size'
     max_page_size = 100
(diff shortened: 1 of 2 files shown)


data_converter/pagination.py
```diff
@@ -8,15 +9,13 @@ class CustomPagination(PageNumberPagination):
     page = DEFAULT_PAGE
```
roman-tiukh, line 9: what is the purpose of the variable `DEFAULT_PAGE`?

data_converter/pagination.py
```diff
@@ -8,15 +9,13 @@ class CustomPagination(PageNumberPagination):
...
-            },
-            # 'total': self.page.paginator.count,
-            'page': int(self.request.GET.get('page', DEFAULT_PAGE)),
-            'page_size': int(self.request.GET.get('page_size', self.page_size)),
-            'results': data
+            'count': self.page.paginator.count,
+            'last_page': math.ceil(self.request.GET.get(
+                'last_page',
+                self.page.paginator.count/int(self.request.GET.get('page_size', self.page_size))
+            )),
```
roman-tiukh, line 19: why `self.request.GET.get` here? at 16 line

roman-tiukh: hardcode page_size you can replace with class variable

data_converter/pagination.py
Baranovsky1986: deleted it

data_converter/pagination.py
Baranovsky1986: fixed according your comments

PR #191, opened 2020-06-26
Custom User Registration & Reset Passowrd

PyDev777:


users/views.py
```diff
@@ -1,17 +1,141 @@
...
+            return Response(DataOceanUserSerializer(request.user).data, status=200)
         return Response(status=403)
+
+
+class CustomRegisterView(views.APIView):
+    permission_classes = (AllowAny, )
+
+    def post(self, request: Request, *args, **kwargs):
+
+        # отримати і перевірити дані
```
Anntroy, line 37: We agreed to do all comments in English

PyDev777: Fixed

PR #196, opened 2020-07-03
Filters

Baranovsky1986:
add  file filters with class RegisterFilter

### DIFF
--- a/data_ocean/filters.py
+++ b/data_ocean/filters.py
@@ -1,18 +1,9 @@
 import django_filters
-from data_ocean.models import Register
-
 
 class RegisterFilter(django_filters.FilterSet):
+    source_last_update = django_filters.DateFromToRangeFilter()
     o = django_filters.OrderingFilter(
         fields=(
             ('id', 'name', 'source_last_update')
         )
     )
-
-    class Meta:
-        model = Register
-        fields = {
-            'name': ['icontains'],
-            'source_register_id': ['exact'],
-            'source_last_update': ['exact', 'gt', 'lt'],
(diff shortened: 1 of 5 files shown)


data_ocean/filters.py
```diff
@@ -0,0 +1,18 @@
...
+        fields=(
+            ('id', 'name', 'source_last_update')
+        )
+    )
+
+    class Meta:
+        model = Register
+        fields = {
+            'name': ['icontains'],
+            'source_register_id': ['exact'],
```
roman-tiukh, line 16: SearchFilter, we need SearchFilter

data_ocean/filters.py
```diff
@@ -0,0 +1,18 @@
...
+            'source_last_update': ['exact', 'gt', 'lt'],
```
roman-tiukh, line 17: we dont need datetime filter, we need date filter...

roman-tiukh: use DateFromToRangeFilter and DateFilter

data_ocean/filters.py
roman-tiukh: https://www.django-rest-framework.org/api-guide/filtering/#searchfilter

Baranovsky1986: added SearchFilter

data_ocean/filters.py
Baranovsky1986: used DateFromToRangeFilter

Baranovsky1986:
> Any requirements??

added django-filter to requirements

PR #212, opened 2020-07-15
Add fields to serializer

PyDev777:


IegorSoboliev:
we do have two similar PRs now))?

PR #213, opened 2020-07-16
company id, company filters

roman-tiukh:


business_register/filters.py
```diff
@@ -0,0 +1,13 @@
+from django_filters import rest_framework as filters
+from .models.company_models import Company
+
+
+class CompanyFilterSet(filters.FilterSet):
+    name = filters.CharFilter(lookup_expr='icontains')
+    address = filters.CharFilter(lookup_expr='icontains')
```
IegorSoboliev, line 7: please, explain, what is the reason to filter address field?

PR #220, opened 2020-07-18
company serializers

roman-tiukh:


business_register/serializers/company_serializers.py
```diff
@@ -54,6 +54,21 @@ class Meta:
         fields = ('op_date', 'reason', 'sbj_state', 'signer_name', 'creditor_reg_end_date')
 
 
+class CompanyShortSerializer(serializers.ModelSerializer):
+    company_type = serializers.StringRelatedField()
+    status = serializers.StringRelatedField()
+    founder_of_count = serializers.SerializerMethodField()
+
+    def get_founder_of_count(self, company):
+        return FounderFull.objects.filter(edrpou=company.edrpou).count()
```
IegorSoboliev, line 63: @roman-tiukh what is the reason to get the only amount of founded companies? Then we should build a new query for having all these companies? We have agreed to get a list and use the len(list) method

roman-tiukh: its faster and no recursion errors

PR #221, opened 2020-07-18
fix founder_of_count

roman-tiukh:


business_register/serializers/company_serializers.py
```diff
@@ -65,7 +65,8 @@ def get_founder_of_count(self, company):
     class Meta:
         model = Company
         fields = (
-            'id', 'name', 'short_name', 'company_type', 'edrpou', 'status',
+            'id', 'name', 'short_name', 'company_type',
```
IegorSoboliev, line 68: @roman-tiukh I think we can omit the 'short_name' field here

roman-tiukh: it does not hurt

PR #224, opened 2020-07-21
Koatuu models

IegorSoboliev:


data_ocean/converter.py
```diff
@@ -19,6 +19,11 @@ class Converter:
     DOWNLOAD_FOLDER = "download/"  # folder to downloaded files
     URLS_DICT = {}  # control remote dataset files update
 
+
+    def load_json(self, json_file):
+        with open(json_file):
+            return json.load(json_file)
```
roman-tiukh, line 25: this work?

data_ocean/models.py
```diff
@@ -8,6 +9,9 @@ class DataOceanModel(models.Model):
     class Meta:
         abstract = True
 
+    def soft_delete(self):
+        self.created_at = now
```
roman-tiukh, line 13: 1. as far as I remember `auto_now_add` will not allow you to set your own value for the `created_at` field.
2. I think you need to call `now()` for get now value.

data_ocean/converter.py
roman-tiukh:
```suggestion
        with open(json_file) as file:
            data = json.load(file)
        return data
```

IegorSoboliev: refactored, thank you!

data_ocean/models.py
IegorSoboliev: sorry, my fault. I meant another field((

IegorSoboliev: fixed

PR #225, opened 2020-07-21
Kved Filters

roman-tiukh:


business_register/views/company_views.py
```diff
@@ -9,7 +9,10 @@
 
 class CompanyView(CachedViewMixin, viewsets.ReadOnlyModelViewSet):
     permission_classes = [AllowAny]
-    queryset = Company.objects.all()
+    queryset = Company.objects.select_related(
+        'parent', 'status', 'bylaw', 'company_type',
```
IegorSoboliev, line 13: My proposition - always show these fields: ['id', 'name', 'edrpou', 'authorized_capital', 'address', 'status']

PR #233, opened 2020-07-23
Made verbose names for all models

IegorSoboliev:


IegorSoboliev:
please, let me know if such naming is ok - then I will continue with other models and the migration

PR #261, opened 2020-08-05
Move is_closed to company property

roman-tiukh:


business_register/models/company_models.py
```diff
@@ -40,6 +40,10 @@ class Company(DataOceanModel):  # constraint for not null in both name & short_n
     code = models.CharField(max_length=510)
     history = HistoricalRecords()
 
+    @property
+    def is_closed(self):
+        return self.status.name in ('припинено', 'в стані припинення',)
```
IegorSoboliev, line 45: @roman-tiukh, 'в стані припинення' is not closed)

roman-tiukh: accepted, I will redo

PR #266, opened 2020-08-12
remove prefetch_related from qs

roman-tiukh:
@IegorSoboliev According to my research, prefetch_related can only be used with models where there is little data


business_register/views/pep_views.py
```diff
@@ -10,7 +10,7 @@
 
 class PepViewSet(CachedViewMixin, viewsets.ReadOnlyModelViewSet):
     permission_classes = [AllowAny]
-    queryset = Pep.objects.prefetch_related('related_persons', 'related_companies').all()
+    queryset = Pep.objects.all()
```
IegorSoboliev, line 13: @roman-tiukh please, explain, what is the reason?

PR #285, opened 2020-08-31
PEP automatic updating with a Celery & db log

PyDev777:


business_register/tasks.py
```diff
@@ -11,10 +10,9 @@ def update_pep():
...
     print('********************')
 
-    file_path = PepDownloader().download()
-    if file_path:
-        print(f'file_path = {file_path}')
-        PepConverter().save_to_db(file_path)
+    file_path, log_id = PepDownloader().download()
+    if file_path and log_id:
+        PepConverter().save_to_db(file_path, log_id)
         PepDownloader().remove_downloaded_file(file_path)
```
roman-tiukh, line 16: why you initialize PepDownloader 2 times?

business_register/converter/pep.py
```diff
@@ -156,7 +158,17 @@ def save_or_update_pep_related_persons(self, related_persons_list, pep):
...
+    def save_to_db(self, json_file, log_id=None):
+
+        if log_id:
+            print(f'save_to_db: log_id = {log_id}.')
+            upd_obj = RegistryUpdaterModel.objects.filter(id=log_id).first()
+            if not upd_obj:
+                print(f"Can't find log record id={log_id} in db!")
+                return
+            upd_obj.update_start = timezone.now()
+            upd_obj.save()
```
roman-tiukh, line 170: this needs to be moved out of save_to_db and the log_id parameter removed

data_ocean/downloader.py
```diff
@@ -38,60 +39,60 @@ def remove_downloaded_file(self, file_path):
             os.remove(file_path)
 
     def download(self):
-        start_time = datetime.now()
+        start_time = timezone.now()
         time_stamp = f'{start_time.date()}_{start_time.hour:02}-{start_time.minute:02}'
-
-        # Create db record & set status "started"
         file_path = f'{self.local_path}{self.reg_name}_{time_stamp}'
```
roman-tiukh, line 44: this needs to be moved into an abstract method

data_ocean/downloader.py
```diff
@@ -38,60 +39,60 @@ def remove_downloaded_file(self, file_path):
...
 
+        upd_obj = RegistryUpdaterModel.objects.create(registry_name=self.reg_name)
+        if not upd_obj:
+            print(f"Can't create log record in db!")
+            return None, None
```
roman-tiukh, line 49: in what cases the object may not be created and the value will be None?

data_ocean/downloader.py
roman-tiukh: its can be saved to instance in file_path attribute

roman-tiukh (changes requested):

roman-tiukh (approved):

PR #292, opened 2020-09-04
Closed free access to our API

IegorSoboliev:


roman-tiukh (changes requested):
This PR will break the visualization "PEP", I will block it for now until figuring out what to do with the visualization.

roman-tiukh (approved):

PR #293, opened 2020-09-07
Added name_eng field to Register model and its values to the fixture

IegorSoboliev:


roman-tiukh (changes requested):
1. need to be added to the sterilizer
2. can you add now fop-register and pep-register to this list?

data_ocean/fixtures/register.json
```diff
@@ -1,74 +1,101 @@
...
-      "data_ocean_list": "/api/company/",
-      "data_ocean_retrieve": "/api/company/{id}/",
+      "list": [
+        "/api/company/",
+        "/api/fop/"
+      ],
+      "retrieve": [
+        "/api/company/{id}/",
+        "/api/fop/{id}/"
+      ],
```
roman-tiukh, line 34: this is not working code.
I propose to make an additional register with FOPs

roman-tiukh (changes requested):

roman-tiukh (approved):

PR #296, opened 2020-09-09
Made EndPoint model, serializer and the fixture

IegorSoboliev:


data_ocean/fixtures/register.json
```diff
@@ -8,13 +8,23 @@
...
       "source_name": "pep.org.ua",
       "source_register_id": null,
-      "list": "/api/pep/",
       "retrieve": "/api/pep/{id}/",
       "url_address": "https://pep.org.ua/uk/data",
       "api_address": "https://pep.org.ua/opendata/persons/json",
       "source_last_update": "2020-09-07"
     }
   },
+  {
```
IegorSoboliev, line 17: Is that a good idea to put it here?

roman-tiukh: its ok.

roman-tiukh: 1. Do it with all registers and all endpoints.
2. Then remove the list and retrieve fields from the Register model.
3. I think that the name_eng field is also needed.

roman-tiukh (changes requested):

data_ocean/fixtures/register.json
IegorSoboliev: done

PyDev777 (approved):

roman-tiukh (approved):

PR #318, opened 2020-10-12
Unified the country filter for the Company list

IegorSoboliev:


business_register/views/company_views.py
```diff
@@ -34,7 +32,7 @@ def get_serializer_class(self):
 
     def get_queryset(self):
         if self.action == 'list':
-            return self.queryset.exclude(from_antac_only=True)
+            return self.queryset.exclude(from_antac_only=True, country_id=None)
```
roman-tiukh, line 35: it won't work as expected...

roman-tiukh (approved):

roman-tiukh (changes requested):
need to change

business_register/views/company_views.py
roman-tiukh, line 35:
```suggestion
            return self.queryset.exclude(from_antac_only=True).filter(country_id__isnull=True)
```

roman-tiukh (changes requested):

roman-tiukh (approved):

PR #325, opened 2020-10-21
Added filter and DB index for Pep code

IegorSoboliev:


roman-tiukh (changes requested):
this should be implemented via @action in the viewset, we talked about this

business_register/views/pep_views.py
```diff
@@ -9,6 +13,7 @@
 
 
 class PepViewSet(CachedViewMixin, viewsets.ReadOnlyModelViewSet):
+    permission_classes = [AllowAny]
```
roman-tiukh, line 16: you miss this line

business_register/views/pep_views.py
```diff
@@ -1,6 +1,10 @@
 from django_filters.rest_framework import DjangoFilterBackend
+from django.shortcuts import get_object_or_404
+from rest_framework.response import Response
 from rest_framework import viewsets
 from rest_framework.filters import SearchFilter
+from rest_framework.decorators import action
+from rest_framework.permissions import AllowAny
```
roman-tiukh, line 7: and this line

business_register/models/pep_models.py
```diff
@@ -42,11 +42,14 @@ class Pep(DataOceanModel):
...
     is_dead = models.BooleanField('помер', default=False)
-    termination_date = models.CharField('дата припинення статусу публічного діяча', max_length=10, null=True)
+    termination_date = models.CharField('дата припинення статусу публічного діяча', max_length=10,
+                                        null=True)
     reason_of_termination = models.CharField('причина припинення статусу публічного діяча',
                                              max_length=125, null=True)
     reason_of_termination_eng = models.CharField('причина припинення статусу публічного діяча англійською',
                                                  max_length=125, null=True)
+    source_id = models.PositiveIntegerField("id from original ANTAC`s DB", unique=True,
+                                            db_index=True, null=True, blank=True)
```
roman-tiukh, line 52: null=True? Do we have peps without source_id?
It is very easy to remove `null=True` in the future.

business_register/views/pep_views.py
```diff
@@ -22,3 +27,9 @@ def get_serializer_class(self):
         if self.action == 'retrieve':
             return PepDetailSerializer
         return super().get_serializer_class()
+
+    @action(methods=['get'], detail=True)
+    def retrieve_by_source_id(self, request, source_id):
```
roman-tiukh, line 32: Nice but method name is very long. Request will be:
```
/api/pep/retrieve_by_source_id/123/
```
if you can make this shortest it will be nice, like
```
/api/pep/source_id/123/
```

roman-tiukh (changes requested):

business_register/views/pep_views.py
```diff
@@ -22,3 +27,9 @@ def get_serializer_class(self):
...
+        pep = get_object_or_404(self.queryset, source_id=source_id)
```
roman-tiukh, line 33: ohh i missed this

use self.get_queryset() instead of self.queryset
```suggestion
        pep = get_object_or_404(self.get_queryset(), source_id=source_id)
```

roman-tiukh (changes requested):

business_register/models/pep_models.py
IegorSoboliev: We do have DB already. Could you explain, what is the way to provide more than 40 k unique values?

roman-tiukh: For this problem, we need to write our own migration.
I don't think there is time for that now. Then just remove the `blank = True`, this will prevent the creation of models without this field.

If in the future we have problems with this, then we will write a migration that first fills in all the blanks and then makes the field `null = False`

business_register/views/pep_views.py
IegorSoboliev: done

business_register/views/pep_views.py
IegorSoboliev: added url_path

business_register/models/pep_models.py
roman-tiukh, line 52: yesterday I asked to leave `blank=False` here, but okay, approved

roman-tiukh (approved):

PR #338, opened 2020-11-02
Project subscription methods

IegorSoboliev:


payment_system/models.py
```diff
@@ -6,17 +6,40 @@
...
+                expiring_date=timezone.localdate() + timezone.timedelta(days=30)
+            )
+        return project_subscription
+
+    @classmethod
+    def disable(cls, project_subscription):
+        project_subscription.status = ProjectSubscription.PAST
+        project_subscription.expiring_date = None
+        project_subscription.save()
+        return project_subscription
```
roman-tiukh, line 41: this is not a classmethod, project_subscription its `self` in instance method

payment_system/serializers/project_subscription_serializers.py
```diff
@@ -0,0 +1,16 @@
...
+from payment_system.models import ProjectSubscription
+
+
+class ProjectSubscriptionSerializer(serializers.ModelSerializer):
+
+    def create(self, validated_data):
+        return ProjectSubscription.create(**validated_data)
+
+    def update(self, instance, validated_data):
+        return ProjectSubscription.disable(instance)
```
roman-tiukh, line 11: No... Its bad, update method is used to update the fields of an object, to deactivate the project, deactivate it inside the view inherited from APIView. In this case, no serializer is needed.

payment_system/models.py
```diff
@@ -6,17 +6,40 @@
...
+        project_subscription = ProjectSubscription.objects.filter(
+            project=project, subscription=subscription).first()
+        if not project_subscription:
+            return ProjectSubscription.objects.create(
+                project=project,
+                subscription=subscription,
+                status=ProjectSubscription.ACTIVE,
+                expiring_date=timezone.localdate() + timezone.timedelta(days=30)
+            )
+        return project_subscription
```
roman-tiukh, line 34: no need to build conditions here, just create a unique_together condition and when creating a uniqueness error, no need to return an object, we must inform users or the frontend about errors that occur.
Although if there can be several subscriptions, then the condition should not be at all

roman-tiukh (changes requested):

payment_system/models.py
IegorSoboliev: done

payment_system/serializers/project_subscription_serializers.py
IegorSoboliev: done

payment_system/models.py
IegorSoboliev: Well, I guess the user should have a right to add a subscription he did use in the past

roman-tiukh (approved):

PR #339, opened 2020-11-04
Added role field to DataOceanUser

IegorSoboliev:


users/models.py
```diff
@@ -43,6 +43,12 @@ def create_superuser(self, email, password, **extra_fields):
 
 
 class DataOceanUser(AbstractUser):
+    PROJECT_INITIATOR = 'project_initiator'
+    PROJECT_PARTICIPANT = 'project_participant'
+    ROLES = [
+        ('PROJECT_INITIATOR', 'project_initiator'),
+        ('PROJECT_PARTICIPANT', 'project_participant'),
+    ]
```
roman-tiukh, line 51: 1. owner
2. member

---

Display names may be more readable...
project_initiator -> Owner
project_participant -> Member

-------------------------------------------------
`('PROJECT_INITIATOR', 'project_initiator'),` - here you dont pass variable, you pass simple string...

users/models.py
```diff
@@ -55,6 +61,7 @@ class DataOceanUser(AbstractUser):
         default=settings.LANGUAGE_CODE,
         blank=True,
     )
+    role = models.CharField(choices=ROLES, max_length=20, default='PROJECT_PARTICIPANT')
```
roman-tiukh, line 64: `default='PROJECT_PARTICIPANT'` -> `default=PROJECT_PARTICIPANT`

users/models.py
roman-tiukh, line 51: read documentation about choices here pls
https://docs.djangoproject.com/en/3.1/ref/models/fields/#choices

roman-tiukh (changes requested):
not in DataOceanUser...
User can be in many projects with different roles...
Do this in User2Project (m2m) model

users/models.py
IegorSoboliev: done, thank you!

users/models.py
IegorSoboliev: Fixed

roman-tiukh (changes requested):

payment_system/models.py
```diff
@@ -5,6 +5,23 @@
...
+    PARTICIPANT = 'participant'
+    ROLES = [
+        ('INITIATOR', 'initiator'),
+        ('PARTICIPANT', 'participant'),
+    ]
+
+    user = models.ForeignKey(settings.AUTH_USER_MODEL, on_delete=models.CASCADE,
+                             related_name='user_projects')
+    project = models.ForeignKey('Project', on_delete=models.CASCADE, related_name='user_projects')
+    role = models.CharField(choices=ROLES, max_length=20, default=PARTICIPANT)
```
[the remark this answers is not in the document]
IegorSoboliev: Do you think, we can have a blank role?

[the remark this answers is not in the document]
IegorSoboliev: added

roman-tiukh (approved):

PR #353, opened 2020-11-18
hotfix: remove debug_toolbar

roman-tiukh:


PyDev777 (approved):

criticus:
@roman-tiukh why was this removed? Please add an explanation to the description

roman-tiukh:
because this library is not specified in the requirements, it is also disabled in INSTALLED_APPS, therefore it generates a fatal error

IegorSoboliev:
As I understand we need it

PR #361, opened 2020-11-26
DO-723 add_users_columns_to_admin_panel

Iva-khar:
DO-723 add_users_columns_to_admin_panel

### DIFF
--- a/users/admin.py
+++ b/users/admin.py
@@ -21,8 +21,10 @@ class DataOceanUserChangeForm(UserChangeForm):
 class DataOceanUserAdmin(UserAdmin):
     add_form = DataOceanUserCreationForm
     form = DataOceanUserChangeForm
-    list_display = ('last_name', 'first_name', 'email', 'is_staff', 'is_active',)
-    list_filter = ('last_name', 'first_name', 'email', 'is_staff', 'is_active',)
+    list_display = ('last_name', 'first_name', 'email', 'date_joined', 'last_login', 'date_of_birth', 'organization',
+                    'position', 'language', 'is_staff', 'is_active', 'is_superuser')
+    list_filter = ('last_name', 'first_name', 'email', 'date_joined', 'last_login', 'date_of_birth', 'organization',
+                   'position', 'language', 'is_staff', 'is_active', 'is_superuser')
     fieldsets = (
         (None, {'fields': ('last_name', 'first_name', 'email', 'password')}),
         ('Permissions', {'fields': ('is_staff', 'is_active')}),
@@ -35,4 +37,3 @@ class DataOceanUserAdmin(UserAdmin):
     )
     search_fields = ('email',)
     ordering = ('email',)
-
--- a/users/admin.py
+++ b/users/admin.py
@@ -29,6 +29,12 @@ class DataOceanUserAdmin(UserAdmin):
                            'organization', 'position', 'language', 'password')}),
         ('Permissions', {'fields': ('is_staff', 'is_active', 'is_superuser')}),
     )
+    add_fieldsets = (
+        (None, {
+            'classes': ('wide',),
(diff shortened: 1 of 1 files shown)


users/admin.py
```diff
@@ -21,8 +21,10 @@ class Meta:
 class DataOceanUserAdmin(UserAdmin):
     add_form = DataOceanUserCreationForm
     form = DataOceanUserChangeForm
-    list_display = ('last_name', 'first_name', 'email', 'is_staff', 'is_active',)
-    list_filter = ('last_name', 'first_name', 'email', 'is_staff', 'is_active',)
+    list_display = ('last_name', 'first_name', 'email', 'date_joined', 'last_login', 'date_of_birth', 'organization',
+                    'position', 'language', 'is_staff', 'is_active', 'is_superuser')
```
roman-tiukh, line 25: do all these fields really need to be displayed in a tabular view?

users/admin.py
```diff
@@ -21,8 +21,10 @@ class Meta:
...
+    list_filter = ('last_name', 'first_name', 'email', 'date_joined', 'last_login', 'date_of_birth', 'organization',
+                   'position', 'language', 'is_staff', 'is_active', 'is_superuser')
```
roman-tiukh, line 27: it is not necessary to add all the fields to the list_filter, only the necessary ones, for text fields there is a search_fields option

roman-tiukh (changes requested):

Iva-khar:
View User page changed to 'read only' mode

users/admin.py
Iva-khar: - removed 'position' and 'is_superuser' fields from **list_display**;
- added 'id', 'date_joined', 'last_login', 'date_of_birth',  'organization', 'position', 'language', 'is_superuser' - all the fields from DB to **fieldsets**;
- removed unused **add_fieldsers** property

users/admin.py
Iva-khar: - removed all text fields except 'language' from **list_filter**;
- added 'last_name', 'first_name', 'organization' to **search_fields**;
- ordering changed to "by date_joined"

users/admin.py
```diff
@@ -21,18 +21,22 @@ class Meta:
...
+        (None, {'fields': ('id', 'last_name', 'first_name', 'email', 'date_joined', 'last_login', 'date_of_birth',
+                           'organization', 'position', 'language', 'password')}),
+        ('Permissions', {'fields': ('is_staff', 'is_active', 'is_superuser')}),
     )
-    add_fieldsets = (
-        (None, {
-            'classes': ('wide',),
-            'fields': ('email', 'password1', 'password2', 'is_staff', 'is_active')}
-         ),
-    )
```
roman-tiukh, line 35: it may still be useful in the future

roman-tiukh (changes requested):

users/admin.py
Iva-khar: I thought it uses only with 'edit/add_new' options. Will return.

roman-tiukh (approved):

users/admin.py
```diff
@@ -21,18 +21,28 @@ class Meta:
...
             'classes': ('wide',),
             'fields': ('email', 'password1', 'password2', 'is_staff', 'is_active')}
          ),
     )
-    search_fields = ('email',)
-    ordering = ('email',)
+    search_fields = ('last_name', 'first_name', 'email', 'organization')
+    ordering = ('date_joined',)
 
+    def has_change_permission(self, request, obj=None):
```
criticus, line 41: please explain why it always returns False

roman-tiukh: if we need to be able to change user data from the admin area, then we can add certain conditions.
Otherwise, this entry disables the ability to edit users from the admin panel.

PR #368, opened 2020-12-01
[DO-175] Change lookup field for detail historical api

Iva-khar:
[DO-175] Change lookup field for detail historical api froom 'pk' to 'company'

[DO-175]: https://dataoceanus.atlassian.net/browse/DO-175

### DIFF
--- a/business_register/views/company_views.py
+++ b/business_register/views/company_views.py
@@ -44,25 +44,30 @@ class CompanyViewSet(CachedViewMixin, viewsets.ReadOnlyModelViewSet):
 
 
 class HistoricalAssigneeView(CachedViewMixin, viewsets.ReadOnlyModelViewSet):
+    lookup_field = 'company'
     queryset = HistoricalAssignee.objects.all()
     serializer_class = HistoricalAssigneeSerializer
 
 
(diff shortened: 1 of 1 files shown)


roman-tiukh (changes requested):
its wrong way. PK its pk for HistoricalCompany model.
for search by company id please add a filter. Something like this:
```
api/some-endpoint/?company_id=123&version_date_after=2020-10-15
```
by edrpou:
```
api/some-endpoint/?edrpou=12345678&version_date_after=2020-10-15
```

Iva-khar:
> its wrong way. PK its pk for HistoricalCompany model.
> for search by company id please add a filter. Something like this:
> 
> ```
> api/some-endpoint/?company_id=123&version_date_after=2020-10-15
> ```
> 
> by edrpou:
> 
> ```
> api/some-endpoint/?edrpou=12345678&version_date_after=2020-10-15
> ```

I don't use HistoricalCompany PK. I'm using 'company' field in Historical Models for search items. Now is't look like: 

`http://localhost:8000/api/historical-signer/{company}/`

It;s wrong cause {company} is string company name. I don't know how to change it to company ID.

About EDRPOU. This field exists in HistoricalCompany Model only. In others company related historical models  we have only foreign key to Company model.

PR #369, opened 2020-12-02
[DO-747] [DO-705] counters, project-token auth for registers

roman-tiukh:


payment_system/migrations/0016_auto_20201201_1802.py
```diff
@@ -0,0 +1,47 @@
+# Generated by Django 3.0.7 on 2020-12-01 18:02
+
+from django.db import migrations, models
+
+
+def set_requests_left(apps, schema):
```
IegorSoboliev, line 6: Consider @property decorator here?

IegorSoboliev (approved):

payment_system/migrations/0016_auto_20201201_1802.py
roman-tiukh: this is a migration, it is not appropriate here

PR #371, opened 2020-12-04
[DO-746] [DO-786] project_token, start of finish of payment_system

roman-tiukh:


payment_system/models.py
```diff
@@ -253,9 +265,9 @@ class Invoice(DataOceanModel):
         help_text='This operation is irreversible, you cannot '
                   'cancel the payment of the subscription for the project.'
     )
-    project_subscription = models.ForeignKey(
+    project_subscription = models.OneToOneField(
```
IegorSoboliev, line 268: Can`t we have a new invoice every month for the same project_subscription?

roman-tiukh: Not sure. Requires discussion.

IegorSoboliev (approved):

PR #373, opened 2020-12-12
[DO-791] tasks after presentation

roman-tiukh:


payment_system/admin.py
```diff
@@ -1,8 +1,20 @@
...
+from django.contrib import messages
+from django.utils import timezone
+
 from .models import Subscription, Invoice, ProjectSubscription
 from rangefilter.filter import DateRangeFilter
 
 
+def set_default_subscription(model_admin, request, queryset):
+    if queryset.count() != 1:
+        model_admin.message_user(request, 'Only one subscription can be default', messages.ERROR)
```
IegorSoboliev, line 11: Maybe, we can use our Email Backend here?

roman-tiukh: Nope. It's just message in django-admin interface

IegorSoboliev (approved):

PR #391, opened 2020-12-31
Getting Pep from DB via code not source_id

IegorSoboliev:


business_register/models/pep_models.py
```diff
@@ -76,7 +76,7 @@ class Pep(DataOceanModel):
     reason_of_termination_eng = models.CharField('причина припинення статусу публічного діяча англійською',
                                                  max_length=125, null=True)
     source_id = models.PositiveIntegerField("id from original ANTAC`s DB", unique=True,
-                                            db_index=True, null=True, blank=True)
+                                            null=True, blank=True)
```
roman-tiukh, line 79: we need db_index here too, at least for endpoint /api/pep/123/source-id/

roman-tiukh (changes requested):

roman-tiukh (approved):

PR #393, opened 2021-01-11
Added views for all users in projects

EmmaSamoilova:


stats/views.py
```diff
@@ -114,3 +115,18 @@ def get_data_for_response():
         return {
             'company_count': Fop.objects.count()
         }
+
+class UsersInProjectsView(views.APIView):
+    @staticmethod
```
roman-tiukh, line 120: not static

stats/views.py
```diff
@@ -114,3 +115,18 @@ def get_data_for_response():
...
+    def get(request):
+        user = request.user
+        user_projects = Project.objects.filter(owner=user.id)
+        users_list = []
+        for project in user_projects:
+            users_list += UserProject.objects.filter(project=project).values_list('user')
+        users_count = len(set(users_list))
```
roman-tiukh, line 127: im not sure at `set` works here

stats/views.py
```diff
@@ -114,3 +115,18 @@ def get_data_for_response():
...
         }
+
+class UsersInProjectsView(views.APIView):
+    @staticmethod
+    def get(request):
+        user = request.user
+        user_projects = Project.objects.filter(owner=user.id)
+        users_list = []
+        for project in user_projects:
+            users_list += UserProject.objects.filter(project=project).values_list('user')
```
roman-tiukh, line 126: we need to count users in all projects, not just projects where the user is the owner

roman-tiukh (changes requested):

stats/views.py
```diff
@@ -114,3 +115,12 @@ def get_data_for_response():
         return {
             'company_count': Fop.objects.count()
         }
+
```
roman-tiukh, line 118: PEP8 - 2 blank lines beetwen classes

stats/views.py
```diff
@@ -114,3 +115,12 @@ def get_data_for_response():
...
+class UsersInProjectsView(views.APIView):
+    def get(self, request):
+        user = request.user
+        user_projects = UserProject.objects.filter(user_id=user.id).values_list('project')
```
roman-tiukh, line 122: add `flat=True` for better practice
```suggestion
        user_projects = UserProject.objects.filter(
            user_id=user.id
        ).values_list('project', flat=True)
```

stats/views.py
```diff
@@ -114,3 +115,12 @@ def get_data_for_response():
...
+        users_count = UserProject.objects.filter(project__in=list(user_projects)).order_by('user').distinct('user').count()
```
roman-tiukh, line 123: Exclude self
```suggestion
        users_count = UserProject.objects.filter(
            project__in=list(user_projects),
        ).exclude(user=user).order_by('user').distinct('user').count()
```

roman-tiukh (changes requested):

roman-tiukh (approved):

PR #427, opened 2021-01-25
upgrade model founder

Iva-khar:


business_register/models/company_models.py
```diff
@@ -167,11 +167,14 @@ class Founder(DataOceanModel):
     company = models.ForeignKey(Company, on_delete=models.CASCADE, related_name='founders',
                                 verbose_name='є засновником компанії/організації')
     info = models.CharField('наявні дані', max_length=2015)
+    info_additional = models.CharField('додаткові наявні дані', max_length=2015)
+    info_beneficiary = models.CharField('наявні дані бенефіціара', max_length=2015)
```
roman-tiukh, line 171: @IegorSoboliev do we need to save this information?

IegorSoboliev: This makes sense

Iva-khar: it's  for storing original information

roman-tiukh (approved):

PR #439, opened 2021-02-02
[DO-1163] More stats

IegorSoboliev:


stats/views.py
```diff
@@ -127,3 +128,40 @@ def get(self, request):
...
+            'pep_related_persons_count': Pep.objects.filter(is_pep=False).count()
+        }
+
+
+class PepLinkedCompaniesCountView(WarmedCacheGetAPIView):
+    @staticmethod
+    def get_data_for_response():
+        return CompanyLinkWithPep.objects.aggregate(
+            pep_related_companies_count=Count('company', distinct=True)
+        )
```
roman-tiukh, line 154: maybe json need to be returned?

roman-tiukh (changes requested):

stats/views.py
IegorSoboliev: it does)

IegorSoboliev: image

roman-tiukh (approved):

PR #442, opened 2021-02-02
[DO-1165] Update company type dictionary

Iva-khar:
When creating the new company type in DB the company_type_dictionary doesn't updating. That leads to error when CompanyConverter tries to create new company type in DB again.

### DIFF
--- a/business_register/converter/company_converters/company.py
+++ b/business_register/converter/company_converters/company.py
@@ -151,6 +151,8 @@ class CompanyConverter(BusinessConverter):
 
     def create_company_type(self, name, name_eng):
         company_type = CompanyType.objects.create(name=name, name_eng=name_eng)
+        self.all_ukr_company_type_dict[name] = company_type
+        self.all_eng_company_type_dict[name_eng] = company_type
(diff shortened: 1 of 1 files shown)


business_register/converter/company_converters/company.py
```diff
@@ -162,12 +162,14 @@ def save_or_get_company_type(self, type_from_record, locale):
             if not company_type:
                 name_eng = self.COMPANY_TYPES_UK_EN.get(name)
                 company_type = self.create_company_type(name, name_eng)
+                self.all_ukr_company_type_dict[name] = company_type
```
IegorSoboliev, line 165: I think we can move that to self.create_company_type() method

IegorSoboliev (changes requested):

business_register/converter/company_converters/company.py
Iva-khar: I don't see difference. But why not...

Iva-khar: done

IegorSoboliev (approved):

PR #470, opened 2021-02-09
[DO-1234] New Manager for DataOcean model

IegorSoboliev:


data_ocean/models.py
```diff
@@ -3,11 +3,20 @@
 from django.utils.translation import gettext_lazy as _
 
 
+class DataOceanManager(models.Manager):
+    # exclude soft-deleted objects from queryset
+    def get_queryset(self):
+        return super().get_queryset().exclude(deleted_at__isnull=False)
```
roman-tiukh, line 9: NOT (deleted_at NOT IS NULL) ?
```suggestion
        return super().get_queryset().filter(deleted_at__isnull=True)
```
deleted_at IS NULL

data_ocean/models.py
```diff
@@ -3,11 +3,20 @@
...
+
+
 class DataOceanModel(models.Model):
     created_at = models.DateTimeField(auto_now_add=True)
     updated_at = models.DateTimeField(auto_now=True, null=True, blank=True)
     deleted_at = models.DateTimeField(null=True, blank=True, default=None)
 
+    objects = DataOceanManager()
+    include_deleted_objects = models.Manager()
```
roman-tiukh, line 18: long manager name.
try something like `all_objects`

roman-tiukh (changes requested):

data_ocean/models.py
IegorSoboliev: first query: cost=0.29..21854.25 rows=45605 width=1185
second guery: cost=0.29..21854.25 rows=45605 width=1185

data_ocean/models.py
IegorSoboliev: I think this is more clear name

roman-tiukh (approved):

PR #474, opened 2021-02-10
[DO-1254] Improve format date function

Iva-khar:
The format date function returns error if invalid input now. It should return nothing in that case.

### DIFF
--- a/data_ocean/utils.py
+++ b/data_ocean/utils.py
@@ -64,8 +64,6 @@ def format_date_to_yymmdd(str_ddmmyy):
         if '/' in str_ddmmyy:
             str_ddmmyy = str_ddmmyy.replace("/", ".")
         return datetime.datetime.strptime(str_ddmmyy, "%d.%m.%Y").strftime("%Y-%m-%d")
-    else:
-        return
 
 
 # checking if exists, then converting to string
--- a/data_ocean/utils.py
+++ b/data_ocean/utils.py
@@ -58,11 +58,14 @@ def cut_first_word(string):
 
 # deleting ';' from a string, then converting it to the datetime format 'Y-m-d'
 def format_date_to_yymmdd(str_ddmmyy):
-    if ';' in str_ddmmyy:
-        str_ddmmyy = str_ddmmyy.replace(";", "")
-    if '/' in str_ddmmyy:
-        str_ddmmyy = str_ddmmyy.replace("/", ".")
-    return datetime.datetime.strptime(str_ddmmyy, "%d.%m.%Y").strftime("%Y-%m-%d")
+    if str_ddmmyy:
+        if ';' in str_ddmmyy:
+            str_ddmmyy = str_ddmmyy.replace(";", "")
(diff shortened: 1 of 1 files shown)


data_ocean/utils.py
```diff
@@ -58,11 +58,14 @@ def cut_first_word(string):
...
-        str_ddmmyy = str_ddmmyy.replace("/", ".")
-    return datetime.datetime.strptime(str_ddmmyy, "%d.%m.%Y").strftime("%Y-%m-%d")
+    if str_ddmmyy:
+        if ';' in str_ddmmyy:
+            str_ddmmyy = str_ddmmyy.replace(";", "")
+        if '/' in str_ddmmyy:
+            str_ddmmyy = str_ddmmyy.replace("/", ".")
+        return datetime.datetime.strptime(str_ddmmyy, "%d.%m.%Y").strftime("%Y-%m-%d")
+    else:
+        return
```
roman-tiukh, line 68: else return?

data_ocean/utils.py
```diff
@@ -58,11 +58,14 @@ def cut_first_word(string):
 
 # deleting ';' from a string, then converting it to the datetime format 'Y-m-d'
 def format_date_to_yymmdd(str_ddmmyy):
-    if ';' in str_ddmmyy:
-        str_ddmmyy = str_ddmmyy.replace(";", "")
-    if '/' in str_ddmmyy:
-        str_ddmmyy = str_ddmmyy.replace("/", ".")
-    return datetime.datetime.strptime(str_ddmmyy, "%d.%m.%Y").strftime("%Y-%m-%d")
+    if str_ddmmyy:
```
IegorSoboliev, line 61: Maybe, it is better to make such a check before calling that function?

data_ocean/utils.py
Iva-khar: It used in in like this callings:
```
creditor_reg_end_date = format_date_to_yymmdd(
                record.xpath('TERMINATION_STARTED_INFO/CREDITOR_REQ_END_DATE')[0].text) or '1990-01-01'
```
It's error when invalid input instead '1990-01-01'

Iva-khar: I need just return

data_ocean/utils.py
Iva-khar: Many checkings in different places vs one check in function? Disagree

IegorSoboliev: Hope sometimes we do have data without None values)

data_ocean/utils.py
Iva-khar: If you mean that 
```
else:
    return 
```
is excess?

roman-tiukh: yes

Iva-khar: done

data_ocean/utils.py
Iva-khar: )

roman-tiukh (approved):

PR #490, opened 2021-02-14
[DO-1153]  Fix the issue with an inappropriate number of users

Andrii17s:
Fixed
https://razom.atlassian.net/browse/DO-1153

### DIFF
--- a/payment_system/serializers.py
+++ b/payment_system/serializers.py
@@ -74,11 +74,8 @@ class ProjectListSerializer(serializers.ModelSerializer):
         return obj.user_projects.get(user=self.context['request'].user).status
 
     def get_users_count(self, obj):
-        active = 0
-        for one in obj.users.all():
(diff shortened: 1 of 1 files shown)


payment_system/serializers.py
```diff
@@ -74,7 +74,11 @@ def get_status(self, obj):
         return obj.user_projects.get(user=self.context['request'].user).status
 
     def get_users_count(self, obj):
-        return obj.users.count()
+        active = 0
+        for one in obj.users.all():
+            if obj.user_projects.get(user=one).status == 'active':
+                active += 1
+        return active
```
roman-tiukh, line 81: try this:
```suggestion
        return obj.users.filter(user_project__status=UserProject.ACTIVE).count()
```

roman-tiukh (changes requested):

roman-tiukh (approved):

PR #499, opened 2021-02-18
Added validation for boolean filters

IegorSoboliev:


business_register/filters.py
```diff
@@ -75,7 +76,8 @@ class PepFilterSet(filters.FilterSet):
     fullname = filters.CharFilter(lookup_expr='icontains')
     fullname_transcriptions_eng = filters.CharFilter(lookup_expr='icontains')
     updated_at = filters.DateFromToRangeFilter()
-
+    is_pep = ValidationBooleanFilter()
+    is_dead = ValidationBooleanFilter()
```
roman-tiukh, line 80: remove this fields from `Meta.fields`

data_ocean/filters.py
```diff
@@ -1,7 +1,25 @@
...
 
 
+class ValidationBooleanField(forms.NullBooleanField):
+
+    def validate(self, value):
+        if str(value) not in ('true', 'false', '1', '0', 'True', 'False'):
+            raise ValidationError(
+                {self.label:
+                     _('Invalid request parameter: could be True, False, true, false, 1, 0')}
+            )
```
roman-tiukh, line 16: add current  invalid value for message
```suggestion
            raise ValidationError({
                self.label: _(f'Invalid request parameter: "{value}", could be True, False, true, false, 1, 0'),
            })
```

roman-tiukh (changes requested):

business_register/filters.py
IegorSoboliev: done

data_ocean/filters.py
IegorSoboliev: value in form already is cleaned. So moved the whole process to a custom widget again

roman-tiukh (approved):

PR #510, opened 2021-02-22
[DO-1055] Add to converter the without removing outdated case

Iva-khar:


data_ocean/converter.py
```diff
@@ -246,7 +246,10 @@ def process(self, start_index=0):
         if records_len:
             self.save_to_db(records)
         if start_index == 0:
-            self.delete_outdated_companies()
+            try:
+                self.delete_outdated()
+            except:
+                print('This converter has not ability to remove outdated records')
```
roman-tiukh, line 252: dont catch all exceptions

roman-tiukh (changes requested):

data_ocean/converter.py
roman-tiukh: you must understand what error you are catching

roman-tiukh: also try hasattr or getattr builtins

roman-tiukh (changes requested):

data_ocean/converter.py
Iva-khar: refactored

PR #549, opened 2021-03-04
[DO-1382] Add possibility for User to pay overdue Invoice

Andrii17s:


roman-tiukh (changes requested):
These changes are not enough for the full functioning of the payment system. A call is required to explain the logic of the payment system

PR #553, opened 2021-03-05
fix migrations and converter

Iva-khar:


roman-tiukh (approved):

IegorSoboliev (commented):
There is also problem on 729:             
`                name = Company.objects.filter(edrpou=edrpou, source=Company.GREAT_BRITAIN_REGISTER).first().name or ''
`
This will rename companies from GREAT_BRITAIN_REGISTER with the same edrpous

PR #618, opened 2021-03-23
[ DO-1240 ] add and validate fields to users model updated

Alexandr8854:
add and validate fields to users model updated


Alexandr8854:
@roman-tiukh I tried to load it into another branch

roman-tiukh (approved):

PR #645, opened 2021-03-29
Added file for former pep.prg.ua users

EmmaSamoilova:


roman-tiukh (approved):
no spaces in file names

PR #664, opened 2021-04-05
[DO-1498, 1499, 1614] Added admin, serializer, viewset and url for sanctions models

IegorSoboliev:


business_register/views/sanction_views.py
```diff
@@ -0,0 +1,33 @@
...
+        'name',
+        'object_origin_name',
+        'registration_number',
+        'taxpayer_number',
+        'address',
+        'place_of_birth',
+        'types_of_sanctions__name',
+        'position',
+        'country__name'
+    )
```
roman-tiukh, line 33: too many search fields)

roman-tiukh (approved):

business_register/views/sanction_views.py
IegorSoboliev: ))

PR #676, opened 2021-04-07
DO-1517 Service e-mail correction

Litsyshyn:
Look at this and if you think that is useless, i'll revert commit and ask PM about close the task. https://razom.atlassian.net/browse/DO-1517


roman-tiukh (changes requested):
compilemessages?

roman-tiukh (approved):

PR #685, opened 2021-04-11
[DO - 1417] Add an optional "Phone" field to the registration form, as well as to the profile.

Andrii17s:
https://github.com/3v-workspace/DataOcean/pull/126/


roman-tiukh (changes requested):
Need to update registration views

roman-tiukh (approved):

PR #777, opened 2021-05-13
[DO-1793] Upgrade fop model

Iva-khar:
Add fields for full parsing registration info & storing source data
Typical data format in source:
`REGISTRATION : 01.01.0001; 12.03.2013; 22741750000032578`


roman-tiukh:
@IegorSoboliev.
Interesting... 2 registration dates

business_register/models/fop_models.py
```diff
@@ -11,6 +11,8 @@ class Fop(DataOceanModel):
     address = models.CharField(_('address'), max_length=500, null=True)
     status = models.ForeignKey(Status, on_delete=models.CASCADE, verbose_name=_('status'))
     registration_date = models.DateField(_('registration date'), null=True, blank=True)
+    registration_date_second = models.DateField(_('registration date second'), null=True, blank=True)
```
IegorSoboliev, line 14: Like two dates of birth)). And what is the reason? Reopening business?

IegorSoboliev (approved):

business_register/models/fop_models.py
Iva-khar: After 07-08 2004 this dates are the same. It seems like in 2004 was reregistration

PR #781, opened 2021-05-14
[DO-1798] Fix report message

Iva-khar:


Iva-khar:
Added long_time_converter field

roman-tiukh (approved):

PR #784, opened 2021-05-15
Added field for Ukrainian names to Country model

EmmaSamoilova:


roman-tiukh (changes requested):
let's use suffixes `_en` and `_uk`

IegorSoboliev (approved):

location_register/management/commands/add_country_ukrainian_name.py
```diff
@@ -0,0 +1,276 @@
...
+        '1': ['україна', 'ukraine'],
+        '2': ['австралія', 'australia'],
+        '3': ['австрія', 'austria'],
+        '4': ['азербайджан', 'azerbaijan'],
+        '5': ['аландські острови', 'aland islands'],
+        '6': ['албанія', 'albania'],
+        '7': ['алжир', 'algeria'],
+        '8': ['американське самоа', 'american samoa'],
+        '9': ['американські віргінські острови', 'virgin islands (u.s.)'],
+        '10': ['ангілья', 'anguilla'],
```
roman-tiukh, line 35: ??? why numbers?
```
{
  'ukraine': 'україна',
  ...
}
```

location_register/models/address_models.py
```diff
@@ -1,11 +1,12 @@
...
-from django.utils.translation import ugettext_lazy as _
+from django.utils.translation import gettext_lazy as _
 
 from data_ocean.models import DataOceanModel
 
 
 class Country(DataOceanModel):
-    name = models.CharField(_('name'), max_length=60, unique=True)
+    name = models.CharField(_('english name'), max_length=60, unique=True)
+    name_uk = models.CharField(_('ukrainian name'), max_length=100, default='')
```
roman-tiukh, line 9: unique=True?

roman-tiukh (changes requested):

location_register/management/commands/add_country_ukrainian_name.py
EmmaSamoilova: These numbers are only needed to identify countries in the register of dictations. There is no country name in the dictations register, only their id in the database. To understand what these id mean, Iegor threw me a doc file with an id and a name. In order not to duplicate similar dictionaries twice, I combined them and import this dictionary into the declaration converter.

location_register/models/address_models.py
EmmaSamoilova: default='' will there be an error?

location_register/management/commands/add_country_ukrainian_name.py
IegorSoboliev: I think we need this only in the declaration module

location_register/models/address_models.py
roman-tiukh: in this case we can use `null=True, default=None, unique=True`

roman-tiukh (approved):

location_register/management/commands/add_country_ukrainian_name.py
```diff
@@ -0,0 +1,277 @@
...
+        countries = countries_dict()
+        for country in countries:
+            nacp_id = country
+            country_uk = countries[country][0]
+            country_en = countries[country][1]
+            country, created = Country.objects.update_or_create(
+                name=country_en, defaults={'nacp_id': nacp_id, 'name_uk': country_uk}
+            )
+            if created:
+                logger.info(f'New country {country_en}')
```
roman-tiukh, line 22: self.stdout

roman-tiukh (approved):

PR #804, opened 2021-05-20
[DO-1828] Async fop

Iva-khar:


roman-tiukh (commented):
1. Need to be tested on test-server
1. many duplicated code with fop full converter
1. on servers we have 2 cores

PR #805, opened 2021-05-20
[DO-1466] Add link to pep.org.ua field

Iva-khar:


roman-tiukh (changes requested):
no need to add field to model, just add property and calculate url in runtime

Iva-khar:
> no need to add field to model, just add property and calculate url in runtime

Done. Add to detail, list & short PEP serializers

business_register/models/pep_models.py
```diff
@@ -35,6 +35,7 @@ class Pep(DataOceanModel):
         (PEP_ASSOCIATED_PERSON, _("Associated person")),
         (PEP_FAMILY_MEMBER, _("Family member")),
     )
+    PEP_ORG_UA_PROFILE_URL = 'https://pep.org.ua/uk/person/'
```
roman-tiukh, line 38: no need this var

roman-tiukh (changes requested):

business_register/models/pep_models.py
Iva-khar: deleted

roman-tiukh (approved):

IegorSoboliev (approved):

PR #806, opened 2021-05-20
[DO-1480] Add organization to registration

Iva-khar:


users/serializers.py
```diff
@@ -99,6 +106,7 @@ def validate(self, data):
             'email': 'Email',
             'first_name': _('First Name'),
             'last_name': _('Last Name'),
+            'organization': _('Organization')
```
roman-tiukh, line 109: no need

roman-tiukh (changes requested):

users/serializers.py
Iva-khar: deleted

Iva-khar:
Tested & translated

roman-tiukh (approved):

PR #819, opened 2021-05-26
[DO-1480]Storing Securities 

IegorSoboliev:


business_register/converter/declaration.py
```diff
@@ -214,6 +212,165 @@ def save_income(self, incomes_data, declaration):
...
+            if issuer_address in self.NO_DATA:
+                issuer_address = ''
+            issuer_registration_number = data.get('emitent_ua_company_code')
+            if issuer_registration_number not in self.NO_DATA:
+                issuer = Company.objects.filter(
+                    edrpou=issuer_registration_number,
+                    source=Company.UKRAINE_REGISTER
+                ).first()
+                if not issuer:
+                    logger.error(
```
EmmaSamoilova, line 285: You can use here self.log_error(message)

    def log_error(self, message):        logger.error(f'Declaration id {self.current_declaration.nacp_declaration_id} : {message}')

IegorSoboliev: Good idea, used it

roman-tiukh (approved):

PR #832, opened 2021-06-01
[DO 1777] Test Storing property from Declaration

EmmaSamoilova:


business_register/converter/declaration.py
```diff
@@ -728,6 +732,9 @@ def save_property(self, property_data, declaration):
             'Офіс': Property.OFFICE
         }
         for data in property_data:
+            if type(data['objectType']) == int:
```
roman-tiukh, line 735: better
```suggestion
            if type(data['objectType']) != str:
```

roman-tiukh (changes requested):
must be 1 migration at finish

roman-tiukh (changes requested):
image

PR #839, opened 2021-06-01
[DO-1809] Translate fields from/to relationship types

Andrii17s:


roman-tiukh (changes requested):
you need to create 2 other fields:
from_person_relationship_type_en and to_person_relationship_type_en

roman-tiukh (approved):

PR #841, opened 2021-06-02
[DO-1846]Added submission date to Declaration

IegorSoboliev:


business_register/converter/declaration.py
```diff
@@ -1,5 +1,6 @@
 import logging
 
+from dateutil.parser import isoparse
```
roman-tiukh, line 3: what is dateutil? And why you need this?

roman-tiukh (changes requested):

business_register/converter/declaration.py
IegorSoboliev: sorry, deleted

roman-tiukh (approved):

business_register/converter/declaration.py
roman-tiukh:
```
datetime.fromisoformat('2020-01-28T22:43:30+02:00')
```

PR #862, opened 2021-06-08
DO-1844 Store unfinished construction

OlexandrTopuzov:


OlexandrTopuzov (commented):
@roman-tiukh 
@EmmaSamoilova

business_register/models/declaration_models.py
```diff
@@ -767,6 +767,13 @@ class Property(DataOceanModel):
         null=True,
         help_text=_('valuation of the property')
     )
+    is_unfinished = models.BooleanField(
+        _('is unfinished constractions'),
+        null=True,
+        blank=True,
+        default=None,
+        help_text=_('is unfinished constractions'),
+    )
```
roman-tiukh, line 776: why we need this fied?

roman-tiukh (changes requested):

business_register/models/declaration_models.py
OlexandrTopuzov: so said to do Єhor, I suppose that the structures in the subsequent can beat the completion

roman-tiukh: BooleandField should have default value and dont have null=True

OlexandrTopuzov: the fact is that we need to apply this field only to type 6 and if the default is true, then all types will be true

roman-tiukh:
wait for Iegor`s review

roman-tiukh (approved):

business_register/converter/declaration.py
```diff
@@ -615,7 +615,28 @@ def save_luxury_item(self, luxuries_data, declaration):
 
     # TODO: implement as save_property()
     def save_unfinished_construction(self, unfinished_construction_data, declaration):
-        pass
+        for data in unfinished_construction_data:
+            additional_info = data.get('objectType')
+            # TODO: add country
+            country = self.find_country(data['country'])
```
EmmaSamoilova, line 621: There may be a KeyError if the 'country' key is missing. Or is 'country' in all declarations?

EmmaSamoilova, line 621: There may be a KeyError if the 'country' key is missing. Or is 'country' in all declarations?

PR #877, opened 2021-06-14
[DO-1980] Transliterate place of work

Iva-khar:
Added property to Pep model


EmmaSamoilova (approved):

business_register/models/pep_models.py
```diff
@@ -166,6 +167,13 @@ def place_of_birth_en(self):
         else:
             return None
 
+    @property
+    def last_employer_en(self):
+        if self.last_employer:
+            return transliterate(translate_company_type_in_string(self.last_employer))
+        else:
```
roman-tiukh, line 174: `else` is redundant here

roman-tiukh (approved):

PR #901, opened 2021-06-23
[DO-1860]Storing corporate rights

IegorSoboliev:


roman-tiukh (approved):

business_register/converter/declaration.py
```diff
@@ -680,6 +683,75 @@ def save_income(self, incomes_data, declaration):
...
+                company_type_name=company_type_name,
+                company_registration_number=company_registration_number,
+                country=country,
+                company=company,
+                value=value,
+                share=share,
+                is_transferred=is_transferred
+            )
+
+            acquisition_date = simple_format_date_to_yymmdd(data.get('owningDate'))
```
EmmaSamoilova: in some declarations the date is not in the correct format, which may cause an error

EmmaSamoilova (approved):

business_register/converter/declaration.py
IegorSoboliev: ValueError is caught in this method

PR #907, opened 2021-06-24
[DO-1888] Added method for save bank account info

EmmaSamoilova:


IegorSoboliev (approved):

business_register/converter/declaration.py
```diff
@@ -395,6 +395,68 @@ def save_liability(self, liabilities_data, declaration):
                     creditor_full_name_eng=creditor_full_name_eng,
                     owner=owner)
 
+    def save_bank_account(self, account_data, declaration, pep):
+        all_banks = list(Money.objects.filter(declaration__pep=pep, type=Money.BANK_ACCOUNT).values_list(
+            'bank_registration_number',
+            flat=True))
```
roman-tiukh, line 401: why one more list() convertation? values_list(flat=True) already return list

roman-tiukh (approved):

business_register/converter/declaration.py
EmmaSamoilova: values_list(flat=True) return <QuerySet [1, 2, 3, ...]> and queryset object has no attribute append

PR #908, opened 2021-06-24
[DO - 1720] Create PEP03 scoring rule

Andrii17s:
0)Create BaseScoringRule class
1)Create PEP03 scoring rule

### DIFF
--- a/business_register/pep_scoring/rules.py
+++ b/business_register/pep_scoring/rules.py
@@ -6,6 +6,7 @@ from business_register.models.declaration_models import (Declaration,
                                                          VehicleRight,
                                                          Income,
                                                          Money,
+                                                         PropertyRight,
                                                          )
 from business_register.models.pep_models import (RelatedPersonsLink, Pep)
 from location_register.models.ratu_models import RatuCity
@@ -33,12 +34,13 @@ class IsRealEstateWithoutValue(BaseScoringRule):
             to_person_links__category=RelatedPersonsLink.FAMILY,
         ).values_list('id', flat=True).all()[::1]
         family_ids.append(self.pep.id)
-        have_weight = Property.objects.filter(
-            declaration__pep_id__in=family_ids,
-            valuation__isnull=True,
+        have_weight = PropertyRight.objects.filter(
+            pep_id__in=family_ids,
+            property__valuation__isnull=True,
             type=Property.SUMMER_HOUSE,
-            created_at__year__gte=2015,
-        ).values_list('id', 'declaration_id').all()[::1]
+            acquisition_date__year__gte=2015,
+        ).values_list('property_id', 'property__declaration_id').all()[::1]
+
         if have_weight:
             value = {
                 "weight": 0.4,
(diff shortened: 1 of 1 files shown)


business_register/pep_scoring/rules.py
```diff
@@ -0,0 +1,106 @@
...
+    def calculate_weight(self):
+        family_ids = self.pep.related_persons.filter(
+            to_person_links__category=RelatedPersonsLink.FAMILY,
+        ).values_list('id', flat=True).all()[::1]
+        family_ids.append(self.pep.id)
+        have_weight = Property.objects.filter(
+            declaration__pep_id__in=family_ids,
+            valuation__isnull=True,
+            type=Property.SUMMER_HOUSE,
+            created_at__year__gte=2015,
```
IegorSoboliev, line 40: created_at is the time of storing the record in our DB. We should use PropertyRight.acquisition date field for this check

business_register/pep_scoring/rules.py
```diff
@@ -0,0 +1,106 @@
...
+    def calculate_weight(self):
+        family_ids = self.pep.related_persons.filter(
+            to_person_links__category=RelatedPersonsLink.FAMILY,
+        ).values_list('id', flat=True).all()[::1]
+        family_ids.append(self.pep.id)
+        have_weight = Property.objects.filter(
+            declaration__pep_id__in=family_ids,
+            valuation__isnull=True,
+            type=Property.LAND,
+            created_at__year__gte=2015,
```
IegorSoboliev, line 69: same here

business_register/pep_scoring/rules.py
```diff
@@ -0,0 +1,106 @@
...
+    """
+
+    def calculate_weight(self):
+        family_ids = self.pep.related_persons.filter(
+            to_person_links__category=RelatedPersonsLink.FAMILY,
+        ).values_list('id', flat=True).all()[::1]
+        family_ids.append(self.pep.id)
+        have_weight = Vehicle.objects.filter(
+            declaration__pep_id__in=family_ids,
+            created_at__year__gte=2015,
```
IegorSoboliev, line 96: and here

IegorSoboliev: VehicleRight. acquisition_date

IegorSoboliev: VehicleRight. acquisition_date

IegorSoboliev (changes requested):

business_register/pep_scoring/rules.py
```diff
@@ -0,0 +1,108 @@
...
+        have_weight = PropertyRight.objects.filter(
+            pep_id__in=family_ids,
+            property__valuation__isnull=True,
+            type=Property.SUMMER_HOUSE,
+            acquisition_date__year__gte=2015,
+        ).values_list('property_id', 'property__declaration_id').all()[::1]
+
+        if have_weight:
+            value = {
+                "weight": 0.4,
```
roman-tiukh, line 46: weight should be saved at another field not in json

business_register/pep_scoring/rules.py
```diff
@@ -0,0 +1,108 @@
...
+    Rule 3.1 - PEP03_home
+    weight - 0.4
+    There is no information on the value of the real estate owned by PEP or
+    family members since 2015
+    """
+
+    def calculate_weight(self):
+        family_ids = self.pep.related_persons.filter(
+            to_person_links__category=RelatedPersonsLink.FAMILY,
+        ).values_list('id', flat=True).all()[::1]
```
roman-tiukh, line 35: values_list('id', flat=True) already return list
```suggestion
        family_ids = self.pep.related_persons.filter(
            to_person_links__category=RelatedPersonsLink.FAMILY,
        ).values_list('id', flat=True)
```

IegorSoboliev (approved):

business_register/pep_scoring/rules.py
```diff
@@ -0,0 +1,108 @@
...
+            acquisition_date__year__gte=2015,
+        ).values_list('property_id', 'property__declaration_id')[::1]
+        if have_weight:
+            weight = 0.1
+            value = {
+                "property_id": have_weight[0][0],
+                "declaration_id": have_weight[0][1],
+            }
+            return json.dumps(value), weight
+        return 0
```
roman-tiukh, line 80: 1. no need to call `json.dumps`
2. function should return same result if don't `have_weight` 
like this ->
```suggestion
            return weight, data
        return 0, {}
```

roman-tiukh (approved):

PR #909, opened 2021-06-24
Refactor save vehicle

OlexandrTopuzov:


EmmaSamoilova (approved):

business_register/models/declaration_models.py
```diff
@@ -849,13 +849,15 @@ class Vehicle(DataOceanModel):
     TRUCK = 2
     MOTORBIKE = 3
     BOAT = 4
-    AGRICULTURAL_MACHINERY = 5
+    AIR_MEANS = 5
+    AGRICULTURAL_MACHINERY = 6
```
roman-tiukh, line 853: don`t change numbers for options, allowed only adding new numbers

roman-tiukh (changes requested):

business_register/models/declaration_models.py
OlexandrTopuzov: done

roman-tiukh (approved):

PR #920, opened 2021-06-27
[DO-1755] Create PEP21 scoring rule

Andrii17s:
Rule 21
Monetary assets declared this year exceed the sum of income and amount of monetary assets of the previous year


roman-tiukh (commented):
Do we have good logic for work with currencies?
I find 16 currencies for example:
```
    'UAH',
    'USD',
    'EUR',
    'RUB',
    'CHF',
    'GBP',
    'PLN',
    'CAD',
    'GEL',
    'CNY',
    'HRK',
    'SEK',
    'AUD',
    'TRY',
    'JPY',
    'HUF',
    'CZK',
```

roman-tiukh (changes requested):
blocked for now

business_register/pep_scoring/rules.py
```diff
@@ -104,3 +104,84 @@ def calculate_weight(self):
...
+class IsMoneyFromNowhere(BaseScoringRule):
+    """
+    Rule 21 - PEP21
+    weight - 1.0
+    Monetary assets declared this year exceed the sum of
+    income and amount of monetary assets of the previous year
+    """
+
+    def calculate_weight(self):
+        declarations = Declaration.objects.filter(
```
IegorSoboliev, line 118: I think we have to use only the last two declarations for this rule

Andrii17s: @IegorSoboliev Why two, not just the last one?

PR #935, opened 2021-06-30
[DO-2037]Refactoring code for storing float numbers

IegorSoboliev:


EmmaSamoilova (approved):

business_register/converter/declaration.py
```diff
@@ -86,7 +86,7 @@ def log_error(self, message):
 
     def to_float(self, value):
         if value not in self.NO_DATA:
-            return float(value)
+            return float(value.replace(',', '.'))
```
roman-tiukh, line 89: In some records i finded value like this "37."
```suggestion
            return float(value.replace(',', '.').strip('.,'))
```

roman-tiukh (approved):

PR #941, opened 2021-07-01
Store save rights

OlexandrTopuzov:


business_register/converter/declaration.py
```diff
@@ -1753,15 +1829,15 @@ def save_declaration(self):
...
-                if (
-                        not declaration.spouse
-                        and detailed_declaration_data['step_2']
-                        and not detailed_declaration_data['step_2'].get('isNotApplicable')
-                ):
-                    self.relatives_data = detailed_declaration_data['step_2']['data']
-                    self.save_related_person(pep, declaration)
-                else:
-                    self.relatives_data = None
+                # if (
```
IegorSoboliev, line 1832: we don`t need it no more))). Every method should work)))

business_register/converter/declaration.py
```diff
@@ -505,7 +549,7 @@ def save_liability(self, liabilities_data, declaration):
             if not owner:
                 self.log_error(f'Cannot identify owner of the liability from data({data})')
             else:
-                Liability.objects.create(
+                parent = Liability.objects.create(
```
IegorSoboliev, line 552: maybe we can use a more understandable name like before - 'liability' and so on)

business_register/converter/declaration.py
```diff
@@ -179,6 +179,50 @@ def find_third_person_rights(self, right_data, name_of_rights):
                 )
         return owner_type, full_name, company, company_name
 
+    def save_right(self, parent, acquisition_date, rights_data):
```
IegorSoboliev, line 182: it doesn`t save anything - let`s name it extract_right_data

business_register/converter/declaration.py
```diff
@@ -179,6 +179,50 @@ def find_third_person_rights(self, right_data, name_of_rights):
...
+                    owner_type, full_name, company, company_name = self.find_third_person_rights(data, 'property')
+                elif owner_info.isdigit():
+                    pep = self.find_person(owner_info)
+                    owner_type = BaseRight.FAMILY_MEMBER
+                else:
+                    self.log_error(f'Wrong value for owner_info = {owner_info}. Check property_right ({data})')
+            additional_info = data.get('otherOwnership', '')
+            # TODO: store 'seller', check if this field is only for changes
+            # Possible values = ['Продавець']
+            seller = data.get('seller')
```
IegorSoboliev, line 221: it that a valid task?

business_register/converter/declaration.py
```diff
@@ -179,6 +179,50 @@ def find_third_person_rights(self, right_data, name_of_rights):
...
+
+            return parent, type_right, additional_info, acquisition_date, share, pep, company, full_name, company_name, owner_type
+
+
```
IegorSoboliev, line 225: good job! Consider storing also right from 'person' field

PR #949, opened 2021-07-02
[DO-2104] fixed conversion to float

EmmaSamoilova:


business_register/converter/declaration.py
```diff
@@ -87,7 +87,11 @@ def log_error(self, message):
 
     def to_float(self, value):
         if value not in self.NO_DATA:
-            return float(value.replace(',', '.').strip('.,'))
+            value = value.replace(',', '.').strip('.,')
+            if len(value.split('.')[0]) <= 10:
+                return float(value)
+            else:
+                self.log_error(f'Too many value for valuation = {value}')
```
IegorSoboliev, line 94: We cannot understand the problem from only the value).

IegorSoboliev (approved):

business_register/converter/declaration.py
EmmaSamoilova: done

criticus (approved):

OlexandrTopuzov (approved):

Andrii17s (approved):

business_register/converter/declaration.py
```diff
@@ -1428,58 +1336,7 @@ def save_securities(self, securities_data, declaration):
...
-                acquisition_date=acquisition_date,
-                share=share,
-                pep=pep,
-                company=company,
-                # TODO: decide should we use lower() for storing names
-                full_name=full_name,
-                company_name=company_name,
-                owner_type=owner_type,
-            )
+                self.save_right(securities, data)
```
IegorSoboliev: we should call this method in all cases (for 'person' field also)

business_register/converter/declaration.py
```diff
@@ -1528,52 +1385,7 @@ def save_vehicle(self, vehicles_data, declaration):
...
-                type=type,
-                acquisition_date=acquisition_date,
-                share=share,
-                pep=pep,
-                full_name=full_name,
-                company=company,
-                owner_type=owner_type,
-                company_name=company_name,
-            )
+                self.save_right(vehicle, data)
```
IegorSoboliev: same here and everywhere with a separate right model

PR #962, opened 2021-07-04
[DO - 1719] Create PEP02 scoring rule

Andrii17s:


business_register/pep_scoring/rules.py
```diff
@@ -72,6 +73,50 @@ def save_to_db(self):
     def calculate_weight(self) -> tuple[int or float, dict]:
         pass
 
+@register_rule
+class IsSpendingMore(BaseScoringRule):
+    """
+    Rule 2 - PEP02
+    weight - 0.4
+    The overall value of the property and assets exceeds income 10 or more times for the current year
```
IegorSoboliev, line 81: The overall value of the PROPERTY

PR #963, opened 2021-07-04
[DO - 1727] Create PEP05 scoring rule

Andrii17s:


business_register/pep_scoring/rules.py
```diff
@@ -225,6 +226,64 @@ def calculate_weight(self) -> Tuple[Union[int, float], dict]:
         return 0, {}
 
 
+@register_rule
+class IsGettingRicher(BaseScoringRule):
```
criticus, line 230: class name has to be noun

PR #964, opened 2021-07-04
[DO-1754] Create PEP20 scoring rule

Andrii17s:


business_register/pep_scoring/rules.py
```diff
@@ -164,3 +165,44 @@ def calculate_weight(self) -> tuple[int or float, dict]:
...
+        year = self.declaration.year
+        max_sum = convert_to_usd('UAH', 1500000, year)  # max amount of cash
+        family_ids = self.pep.related_persons.filter(
+            to_person_links__category=RelatedPersonsLink.FAMILY,
+        ).values_list('id', flat=True)[::1]
+        family_ids.append(self.pep.id)
+        declarations = Declaration.objects.filter(
+            year=year,
+            pep_id__in=family_ids,
+        ).values_list('id', flat=True)[::1]
```
IegorSoboliev, line 193: What is the aim of this variable? We don`t need all declarations here

business_register/pep_scoring/rules.py
```diff
@@ -164,3 +165,44 @@ def calculate_weight(self) -> tuple[int or float, dict]:
...
+        assets_USD = 0
+        for declaration_id in declarations:
```
IegorSoboliev, line 195: As we discussed many times, we need hard cash from ONE declaration. The ONE we check

PR #991, opened 2021-07-06
[Contains Migration] Made Income.paid_by_person TextField

IegorSoboliev:


business_register/migrations/0146_auto_20210706_1746.py
```diff
@@ -0,0 +1,18 @@
...
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('business_register', '0145_auto_20210706_1336'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='income',
```
criticus, line 14: is this table empty? How long does the migration run?

PyDev777 (approved):

business_register/migrations/0146_auto_20210706_1746.py
IegorSoboliev: We do have few thousands declarations on our test server. Very quickly)

PR #1002, opened 2021-07-08
fix step_3 rights AttributeError: 'list' object has no attribute 'get'

roman-tiukh:


PyDev777 (approved):

business_register/converter/declaration.py
```diff
@@ -223,7 +223,9 @@ def save_right(self, property, corporate_rights_data):
         rights_data = corporate_rights_data.get('rights')
         if rights_data:
             for right_data in rights_data:
-                type = TYPES.get(right_data.get('ownershipType'))
+                if not isinstance(right_data, dict):
```
IegorSoboliev, line 226: Please, add TODO here to store data that is not in a dict

PR #1054, opened 2021-07-12
Bugfix

Andrii17s:
Fix of this bug without changing model 
https://3v-workspace.slack.com/files/U01GM3B0DPG/F027J9EA0G6/image.png


business_register/converter/declaration.py
```diff
@@ -263,6 +263,9 @@ def save_right(self, property, corporate_rights_data):
 
                     for key, (model_name, first_field) in model_dict.items():
                         if property._meta.model_name == key:
+                            if 'Особа не надала згоду на обробку персональних даних' in full_name:
+                                full_name = 'Особа не надала згоду на обробку персональних даних'
```
IegorSoboliev, line 267: I think full_name also should be '' here

business_register/converter/declaration.py
```diff
@@ -263,6 +263,9 @@ def save_right(self, property, corporate_rights_data):
...
+                                owner_type = BaseRight.NO_INFO_FROM_FAMILY_MEMBER
                             field_dict = {
```
IegorSoboliev, line 269: Let`s establish one more BaseRight.type for these cases?

roman-tiukh (approved):

PR #1061, opened 2021-07-12
[DO-2145] Save empty beneficiaries of

Iva-khar:


criticus:
@Iva-khar what's status of this PR?

PR #1141, opened 2021-07-28
[DO-2077] Add search filter to export xlsx

Iva-khar:


data_ocean/export.py
```diff
@@ -18,9 +20,13 @@
 class ExportToXlsx:
 
     @staticmethod
-    def export(params, export_dict, model_path, filterset_module, serializer_module, user_id):
+    def export(params, export_dict, model_path, filterset_module, serializer_module, view_module, user_id):
         app, model = model_path.split('.')
         queryset = apps.get_model(app, model).objects.all()
+        view = import_string(view_module)
+        request = requests
```
roman-tiukh, line 27: use native request object from django DRF

data_ocean/export.py
```diff
@@ -18,9 +20,13 @@
...
+        setattr(request, 'query_params', params)
+        queryset = SearchFilter.filter_queryset(SearchFilter(), request, queryset, view)
         queryset = import_string(filterset_module)(params,  queryset).qs
         serializer = import_string(serializer_module)(queryset, many=True)
```
roman-tiukh, line 31: you can omit serializer, it redundant

roman-tiukh (changes requested):

data_ocean/export.py
Iva-khar: done

data_ocean/export.py
Iva-khar: I need serializer for `pep_type_display` field, that generates in serializer

roman-tiukh (approved):

PR #1142, opened 2021-07-28
[DO-2248] Add possibility to show information about PEP according to the Full name+DOB and EDRPOU

Andrii17s:
https://razom.atlassian.net/browse/DO-2248


Andrii17s:
> !? @Andrii17s, backend was already done by myself for these features. @Serhii-PM1 we discussed that yesreday

@IegorSoboliev Search by EDRPOU or both are done?

PR #1213, opened 2021-09-08
added id to PropertySerializer

EmmaSamoilova:


roman-tiukh (approved):
maybe you can add id to all serializers one time

EmmaSamoilova:
> maybe you can add id to all serializers one time

Seems to be needed nowhere else

IegorSoboliev (approved):

PR #1228, opened 2021-09-16
[DO-2515]Implemented Person search

IegorSoboliev:


person/views.py
```diff
@@ -23,3 +29,59 @@ class PersonViewSet(RegisterViewMixin,
...
+    def get_serializer_class(self):
+        if self.action == 'retrieve':
+            return PersonDetailSerializer
+        return super().get_serializer_class()
+
+
+class SearchPersonView(generics.ListAPIView):
+    schema = None
+    serializer_class = PersonListSerializer
+    queryset = Person.objects.filter(is_main=True)
```
roman-tiukh, line 42:
```suggestion
    queryset = Person.objects.filter(duplicate_of__isnull=True)
```

person/views.py
```diff
@@ -23,3 +29,59 @@ class PersonViewSet(RegisterViewMixin,
...
+    # TODO: define permission classes
+    permission_classes = [AllowAny, ]
+
+    def list(self, request, *args, **kwargs):
+        last_name_coefficient = 2
+        citizenship_weight = 1
+        last_name_threshold = 0.7
+        last_name = request.query_params.get('last_name')
```
roman-tiukh, line 50: `last_name` should be required.
`first_name`, `middle_name`, `country_name` is optional, so if they are not provided then we don't need annotate

roman-tiukh (changes requested):

person/views.py
IegorSoboliev: As we discussed, last_name will be required on front-end

person/views.py
```diff
@@ -23,3 +29,59 @@ class PersonViewSet(RegisterViewMixin,
...
+            return PersonDetailSerializer
+        return super().get_serializer_class()
+
+
+class SearchPersonView(generics.ListAPIView):
+    schema = None
+    serializer_class = PersonListSerializer
+    queryset = Person.objects.filter(is_main=True)
+    # TODO: define permission classes
+    permission_classes = [AllowAny, ]
```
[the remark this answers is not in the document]
IegorSoboliev: # TODO: define permission classes)

roman-tiukh (changes requested):
validation should be on the backend.
`first_name`, `middle_name`, `country_name` is optional, so if they are not provided then we don't need to annotate

person/views.py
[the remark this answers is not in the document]
roman-tiukh: IsAuthenticated better than AllowAny

person/views.py
IegorSoboliev: added check)

person/views.py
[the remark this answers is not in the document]
IegorSoboliev: changed

person/views.py
```diff
@@ -23,3 +29,67 @@ class PersonViewSet(RegisterViewMixin,
...
+    # TODO: define permission classes
+    permission_classes = [IsAuthenticated, ]
+
+    def list(self, request, *args, **kwargs):
+        # verifying presence of last_name first
+        last_name = request.query_params.get('last_name')
+        if not last_name:
+            return Response(
+                {'first_name': 'This field is required'}
+            )
```
roman-tiukh, line 52:
```suggestion
            return Response(
                {'first_name': 'This field is required'}, status=400
            )
```

roman-tiukh (approved):

EmmaSamoilova (approved):

